PR #149, opened 2019-05-06
Update Changelog

alekksander:
simple typo :)

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -11,7 +11,7 @@ Version 0.5.4:
     * Changed description for offline mode scrobbling for less confusion
     * Fixed scrobbler to not send "playing now" when in offline mode
     * Added common RPM spec file for openSUSE, Fedora, RedHat, CentOS and Mageia
-    * Improved cover manager score system
+    * Improved cover manager store system
     * Added Deezer and Tidal cover providers
     * Added ChartLyrics lyrics provider
     * Added HTTPS workaround for Last.Fm OAuth to solve redirect https to http (localhost) problems


jonaski:
It's not a typo. When you play a song without a album cover it searches for covers on all the cover providers and determiners which one to use based on a score for each image. The same is done when you use fetch all album covers in the cover manager.

The score is calculated based on the following:

* The initial known quality of the provider (Tidal and Deezer are high quality, while Discogs and last.fm are low, musicbrainz is in the middle)
* Whether artist and album have an exact match
* If both artist and album do not match the score will be additionally decreased, so they get a very low score even if they are high resolution.
* The size of the image, higher resolution give a better score
* Whether the image is 1:1 (width and height is identical), the more width and height differ the lower the score will be

Previously there was no score calculation for the artist and album returned by the provider, so a mismatched high resolution cover could get picked over a lower resolution cover that actually matched.
Also the 1:1 score calculation from Clementine was broken.

https://github.com/jonaski/strawberry/blob/63d5018ad6662cae7343128a64fecb0cd707c923/src/covermanager/albumcoverfetchersearch.cpp#L119

https://github.com/jonaski/strawberry/blob/63d5018ad6662cae7343128a64fecb0cd707c923/src/covermanager/albumcoverfetchersearch.cpp#L247

Maybe it could be improved even more, but I've tested it with my 1700 albums and it is very accurate now. Covers from Deezer (1000x1000) and Tidal (1280x1280) are most the time very good and get's picked most of the time, but where artist and album don't match, the others get's picked instead.
For Tidal to work you need to be logged into Tidal with an account, but you don't need a subscription. Deezer works without any login.

PR #196, opened 2019-06-28
Add option: notification on playback resume

m4x3t:
This adds an optional setting to show the notification that is displayed when changing the track,
when resuming playback as well.

While everything works fine for me, there are a few things I'm unsure of.
  * the inconsistency between "playing" and "resumed" for the signal and slot
  * the fact that although the options for a notification on pause and the added on resume look similar,
they do totally different things (one showing only the "Paused" text and the other showing track information)

### DIFF
--- a/src/widgets/osd.cpp
+++ b/src/widgets/osd.cpp
@@ -186,9 +186,9 @@ void OSD::Paused() {
 }
 
 void OSD::Resumed() {
-    if (show_on_resume_) {
-        AlbumArtLoaded(last_song_, last_image_uri_, last_image_);
-    }
+  if (show_on_resume_) {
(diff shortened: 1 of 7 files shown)


src/widgets/osd.cpp
```diff
@@ -183,6 +185,12 @@ void OSD::Paused() {
   }
 }
 
+void OSD::Resumed() {
```
jonaski: Just call it Playing and create a member variable that tracks whether it is paused or not.

m4x3t: I don't really get what you mean with the member variable part, what would the use be?

jonaski: To prevent AlbumArtLoaded() done twice when Playing because it's done now in Resumed() too.

PR #198, opened 2019-06-30
Modify stretch background image functionality

### LINKED ISSUE #177: Background album zoom feature
If possible it would be great that the album cover zooms so that it fills the Playlist fully while keeping ratio. Pleaseeeee :}

m4x3t:
Changes the stretch functionality to fill out
the playlist background completely by zooming
the image (if keep aspect ratio is selected)
instead of filling it only height-wise.

I thought about adding a new zoom option in the settings, but
I think "stretch background image to fill playlist background"
should actually fill the *entire* background instead of what
it did before.
This will resolve #177

### DIFF
--- a/src/playlist/playlistview.cpp
+++ b/src/playlist/playlistview.cpp
@@ -832,8 +832,8 @@ void PlaylistView::paintEvent(QPaintEvent *event) {
             }
           }
           else {
-            int resize_width = qMin(qMin(background_image_.size().width(), (pb_width >= 50 ? (pb_width - 25) : pb_width)), background_image_maxsize_);
-            int resize_height = qMin(qMin(background_image_.size().height(), (pb_height >= 50 ? (pb_height - 25) : pb_height)), background_image_maxsize_);
+            int resize_width = qMin(qMin(background_image_.size().width(), pb_width), background_image_maxsize_);
+            int resize_height = qMin(qMin(background_image_.size().height(), pb_height), background_image_maxsize_);
             cached_scaled_background_image_ = QPixmap::fromImage(background_image_.scaled(resize_width, resize_height, Qt::KeepAspectRatio, Qt::SmoothTransformation));
           }
         }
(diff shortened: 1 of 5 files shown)


jonaski:
I'm fine with adding it as an option, but I don't want it to replace the current fill background, I actually use the stretch setting with keep aspect ratio myself, and the main reason I made it was that I didn't want to cut the image. So I suggest adding an sub option called "Cut image to fit" under the stretch setting

jonaski:
I'm open to suggestions on what to call the different options too.

m4x3t:
How about the other way around, keeping the fill option like it would be with this patch and adding a "Do not cut image" option?

jonaski:
OK I guess

m4x3t:
I've noticed another thing, the height() that is used to stretch is actually the height of the visible playlist viewport + the height of the columns above (artist, album, etc.). This means that a portion of the album art is cut off at the bottom.

gianlucapettinello:
Thanks both for your work and help!

jonaski:
@m4x3t 
I didn't notice it before, but I see that now yeah.
I can fix that if you want, do you want me to merge this now or do you want to fix that too?
The solution is to add a variable on top that calculates the playlist background height
int pb_height = height() - header_->height();
Then replace height() with that.

m4x3t:
I can quickly add that. And thanks, I was just searching for the name of the header.

m4x3t:
Actually, the height() should be replaced by that in the lines below too, where the image positions are calculated, right?

jonaski:
Yes

m4x3t:
Just noticed there's still an error, I need a way to check if the vertical scrollbar is actually visible for the width calculation to be correct.

jonaski:
Nice, now you can remove the - 25 hacks I did.
Just replace (pb_width >= 50 ? (pb_width - 25) : pb_width)) with pb_width
and (pb_height >= 50 ? (pb_height - 25) : pb_height)) with pb_height
Then also remove the - 25 hacks I did for the bottom positions.

jonaski:
Thanks for your contribution.

jonaski:
@m4x3t Feel free to add yourself to the contributors list if you want: https://github.com/jonaski/strawberry/blob/master/src/dialogs/about.cpp

PR #202, opened 2019-07-04
Make playlist ignore articles when sorting

gavinhoward:
This is more correct, and the other way is driving me crazy.

On the note about using `queue2`: when I built with it, not only did it solve the problem I had with `flac` files,  it seemed to have solved my other problems as well, except for this one. I am doing this one first, and then I will start trying to reproduce the problem with `queue2`. I just didn't want to put two different changes into one PR.

### DIFF
--- a/src/collection/collectionmodel.cpp
+++ b/src/collection/collectionmodel.cpp
@@ -1286,14 +1286,12 @@ QString CollectionModel::SortTextForArtist(QString artist) {
 
   artist = SortText(artist);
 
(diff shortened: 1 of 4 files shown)


jonaski:
I think we should do this for artist and album artist too to make it consistent.

gavinhoward:
Hmm...I think you are right. I thought that artists would not have articles in them because they are names, but bands are artists, and they can have articles in the name. Give me a few minutes to get this done.

PR #204, opened 2019-07-05
Switch to queue2 for probe queue

jonaski:
Not sure if this is 100% correct but it works with pause now.
I've tested to pause over night and unpause in the morning, I've also tested to pause many times for serveral hours and unpause works.


gavinhoward:
So far, so good. It's been working for me for several hours.

PR #217, opened 2019-07-17
Use macdeployqt fix #184

### LINKED ISSUE #184: macOS Native widgets on nightly builds
I downloaded the latest nightly build from https://builds.strawbs.org/macos/ and I was surprised to see that the interface doesn't look native at all.

Screenshot 2019-06-20 at 06 46 29

Then I decided to compile the application myself and then all the widgets are native (tabs, scroll bars, buttons etc.)

Screenshot 2019-06-20 at 06 47 25

Do you know why? It's not a deal breaker, just curious to know the reason.

By the way I just realized that the nightly build only takes 123MB in memory and the one I compiled myself following the instructions in your Wiki takes up to 730MB! there must be something wrong obviously :)

Thanks,

MartinDelille:
I removed the old `macdeploy.py` script that might be buggy and replaced it by the Qt official macdeployqt tool (https://doc.qt.io/qt-5/macos-deployment.html)


MartinDelille:
@jonaski If the build is uploaded somewhere, it might be interesting to test before merging.

jonaski:
This is cool, but we need to make sure gstreamer plugins are bundled too, the following list:
https://github.com/jonaski/strawberry/blob/master/dist/macos/macdeploy.py#L57

MartinDelille:
The macdeployqt take care of adding all the plugins/framework/dynamic library in the bundle.

After running it I can see the following:

```
$ ls strawberry.app/Contents/Frameworks
QtConcurrent.framework
QtCore.framework
QtDBus.framework
QtGui.framework
QtMacExtras.framework
QtNetwork.framework
QtPrintSupport.framework
QtSql.framework
QtSvg.framework
QtWidgets.framework
libcdio.19.dylib
libchromaprint.1.dylib
libffi.6.dylib
libfftw3.3.dylib
libgcc_s.1.dylib
libgio-2.0.0.dylib
libglib-2.0.0.dylib
libgmodule-2.0.0.dylib
libgmp.10.dylib
libgnutls.30.dylib
libgobject-2.0.0.dylib
libgstapp-1.0.0.dylib
libgstaudio-1.0.0.dylib
libgstbase-1.0.0.dylib
libgstreamer-1.0.0.dylib
libgsttag-1.0.0.dylib
libhogweed.4.dylib
libintl.8.dylib
libmtp.9.dylib
libnettle.6.dylib
liborc-0.4.0.dylib
libp11-kit.0.dylib
libpcre.1.dylib
libprotobuf.18.dylib
libtasn1.6.dylib
libunistring.2.dylib
libusb-1.0.0.dylib
```

Of couse a test on a non development machine would be great.

MartinDelille:
Ok I have a problem with playback locally...

jonaski:
The gstreamer plugin scanner needs to match the path here:
https://github.com/jonaski/strawberry/blob/master/src/engine/gststartup.cpp#L69
And the plugins need to be in the path below.

All the files here need to be there:
https://github.com/jonaski/strawberry/blob/master/dist/macos/macdeploy.py#L57

The binary won't link to gstreamer plugins so those probably needs to be copied the manually.
So I don't think the qt deploy will know what plugins are needed.
So you could probably strip everthing from the current python deploy script and just keep the gstreamer plugins stuff.

.travis.yml
```diff
@@ -44,8 +44,7 @@ script:
 - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       make -j8 || travis_terminate 1;
       make install || travis_terminate 1;
-      ../dist/macos/macdeploy.py strawberry.app || travis_terminate 1;
-      ../dist/macos/create-dmg.sh strawberry.app;
+      make dmg
```
jonaski: make dmg;

jonaski:
We can merge this into the macos branch first and test it if you fix the issue with make dmg and the conflict

MartinDelille:
It doesn't work: macdeployqt doesn't bundle correctly the gstreamer plugin. I didn't really understood how to combine macdeployqt and macdeploy.py together yet.

jonaski:
@MartinDelille I did it, check the macos branch. I think it works.
The macos branch is uploading to: https://builds.strawbs.net/macos-test/
Test the latest build there and see that there are no errors on the command line and it plays.

jonaski:
I think the problem was that styles/libqmacstyle.dylib was missing, thats fixed by using macdeployqt, it also bundles alot more qt plugins
But macdeployqt did not deploy strawberry-tagreader, the gst-plugin-scanner and gstreamer plugins. So unless we find a way to make macdeployqt handle that too we need to still use the macdeploy.py script too.

MartinDelille:
Unfortunately I tested the build above and it doesn't play sound and doesn't display macos look'n feel. I'll try to understand how to fix this later on.

jonaski:
I think it works for @llucps

jonaski:
I installed macOS in a virtualbox to fix some remaining problems. It works fine for me too with the changes.
Maybe you forgot to compile with cmake .. -DUSE_BUNDLE=ON before you created the dmg, it won't find plugins correctly without that.
The main issue now seem to be performance, the cause might be that something is running in the GUI thread that should be running in it's own thread, there is a lot of thread related errors. Those are not there on Linux or windows.

llucps:
To be fair, the performance is on a par with Clementine.. yes is not the most responsive application ever, but I thought it was because of the QT implementation in macOS.

I only us it mainly to connect with my Airsonic server and Radios in playlists that I curated myself.

Can those GUI threads performance issues be resolved? Is it complicated? I would love to help but I don't know how exactly.

Thanks,

jonaski:
Looking more closely into it, it looks like the two moveToThread()'s errors is related to conflicting libraries, it's picking up QtCore from both /usr/local/Cellar/qt/5.13.0/lib/QtCore.framework/Versions/5/QtCore and /Applications/strawberry.app/Contents/Frameworks/QtCore.framework/Versions/5/QtCore
The moveToThread errors occurs just after, and it only occurs when I use the USE_BUNDLE option. So it's not related to the moveToThread's done in application.cpp (which I initially just assumed it was). So it not the cause for the bad performance.
There is also a lot of these errors, which isn't related to startTimer() and killTimer() calls anywhere in the source code of strawberry (I commented out every use of it just to make sure), but might be related to timers somewhere in the application elsewhere, or incorrect uses of connect's on timers. But finding exactly where is difficult.
```
__logging_message__17:50:27.389 WARN  unknown                          QObject::startTimer: Timers cannot be started from another thread
__logging_message__17:50:27.418 WARN  unknown                          QObject::killTimer: Timers cannot be stopped from another thread
```
But I don't think the performance issues is related to this at all, it might be as you said, Qt performs bad on macOS, but there probably is code that could be better optimized, or some drawing that could be reduced.
I tried to compile with -O3 which should optimize the compilation as much as possible, but didn't notice any significant improvements, (and using that in travis-ci will make the build time a lot slower).

llucps:
It's ok.. thanks to have looked in to it anyway :)

jonaski:
Its using too much cpu time when just idling

image

llucps:
Umm.. strange. With Strawberry open and idle I get 0%

Screenshot 2019-08-19 at 15 29 17

Then playing an mp3 from a m3u file it uses between 5-6% of CPU which is really great compared to Clementine which it took between 10-12% of CPU.

Screenshot 2019-08-19 at 15 27 35

jonaski:
QTimer thread errors were related to the collection watcher, see https://github.com/jonaski/strawberry/issues/324
Fixed now.

PR #220, opened 2019-07-19
Ipod Playlist Support

SamTShaw:
Copy a whole playlist to the ipod and create an entry in Playlists on the iPod

What's not implemented is seeing the playlists on the ipod in the device manager. I'm not sure how that should be displayed, but this is a start to supporting playlists.

### DIFF
--- a/src/device/gpoddevice.cpp
+++ b/src/device/gpoddevice.cpp
@@ -188,6 +188,19 @@ bool GPodDevice::CopyToStorage(const CopyJob &job) {
     return false;
   }
 
+  // Put the track in the playlist, if one is specified
+  if(!job.playlist_.isEmpty()) {
+      // Does the playlist already exist?
+      auto itdbPlaylist = itdb_playlist_by_name(db_, job.playlist_.toUtf8().data());
+      if(itdbPlaylist == nullptr) {
+          // Create the playlist
+          itdbPlaylist = itdb_playlist_new(job.playlist_.toUtf8().data(), false);
+          itdb_playlist_add(db_, itdbPlaylist, -1);
+      }
+      // Playlist should exist so add the track to the playlist
+      itdb_playlist_add_track(itdbPlaylist, track, -1);
+  }
+
   AddTrackToModel(track, url_.path());
 
   // Remove the original if it was requested
--- a/src/device/gpoddevice.cpp
+++ b/src/device/gpoddevice.cpp
@@ -189,16 +189,16 @@ bool GPodDevice::CopyToStorage(const CopyJob &job) {
   }
 
   // Put the track in the playlist, if one is specified
-  if(!job.playlist_.isEmpty()) {
-      // Does the playlist already exist?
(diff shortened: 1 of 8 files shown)


src/device/gpoddevice.cpp
```diff
@@ -182,6 +182,19 @@ bool GPodDevice::CopyToStorage(const CopyJob &job) {
     return false;
   }
 
+  // Put the track in the playlist, if one is specified
+  if(!job.playlist_.isEmpty()) {
```
jonaski: Can you fix the code style here to 
if (!job.playlist_.isEmpty()) {
and the indent to 2 spaces

src/device/gpoddevice.cpp
```diff
@@ -182,6 +182,19 @@ bool GPodDevice::CopyToStorage(const CopyJob &job) {
...
+      // Does the playlist already exist?
+      auto itdbPlaylist = itdb_playlist_by_name(db_, job.playlist_.toUtf8().data());
+      if(itdbPlaylist == nullptr) {
```
jonaski: Here too

src/playlist/playlistlistcontainer.cpp
```diff
@@ -317,6 +326,33 @@ void PlaylistListContainer::ItemDoubleClicked(const QModelIndex &proxy_index) {
 
 }
 
+void PlaylistListContainer::CopyToDevice()
+{
+#ifndef Q_OS_WIN
+    // Reuse the organise dialog, but set the detail about the playlist name
+    if(!organise_dialog_) {
```
jonaski: Here too

jonaski:
There is minor conflict in src/organise/organise.h that needs to be adressed to.

To display playlist, the playlistmanager needs to use another model, look at how the internet model/view is done in Clementine.

SamTShaw:
So I've been having a play around with trying to implement another model for displaying playlists for the device. I might have been going the wrong way though.
I've tried implementing another device class in `DeviceManager`, but that only supports a device being connected and modeled using one device class.

I'm wondering if it's the `CollectionModel` and `CollectionBackend` that is the most appropriate place change to support playlists.

jonaski:
It isn't really easy to do without a lot of work.

I added support for copying album art to the ipod a while ago, but never got it to work.
https://github.com/jonaski/strawberry/blob/master/src/device/gpoddevice.cpp#L155
Does it work for you?
I got a iPod Video 60GB 5th gen, model MA147FB. I don't know if it's supported or not.
I also have a IPOD 80GB 6th gen, model MB029, that one won't read at all, probably not supported by libgpod.

SamTShaw:
I've got an 80GB classic (6th gen), model MB147ZP, that I've been testing these changes with. I haven't checked for album art being copied in strawberry - but I do know libgpod supports it as it's come through from when I was using gtkpod.

jonaski:
I never got gtkpod to work at all, the iPods won't show up at all.
Could you test and see if album covers work for you with strawberry?

PR #229, opened 2019-07-26
Switch to FTS5 with unicode61

jonaski:
I'm looking into switching to FTS5 with the unicode61 tokenizer. This will increase performance and use less memory.

Advantages:
* Require less memory
* Faster
* No custom tokenizer

To do this we need to make sure the search still work like before and be absolutely sure that there aren't any problems, there is no going back.

We also need to make sure distros have SQLite 3.9 or above compiled with SQLITE_ENABLE_FTS5 (or ask packagers to fix it if not).

I've changed the DB file to strawberry-fts5.db so it does not mess with your current DB before we decide to make the switch.

There are no builds of this, so you need to compile yourself. You should also run it from the terminal to see any sql errors.

Just tagging contributors and some others, it will be great if you can test.
Test searching in the collection and see that it works like before.

@MartinDelille @m4x3t @gavinhoward @eclipseo @thomaspierson @SamTShaw @thomaspierson @davidgfnet @alekksander  @lemonzest79  @gianlucapettinello  @renn0xtek9  @FabioLolix 

By the way. I did a hack registering the unicode61 as the unicode tokenizer, this is just to be able to drop the current fts3 tables when upgrading.

To compile you need to switch to the fts5 branch.

```
git clone https://github.com/jonaski/strawberry strawberry-fts5
cd strawberry-fts5
git checkout fts5
mkdir build
cd build
cmake ..
make -j4

```


jonaski:
Codacy doesn't understand virtual tables, thats the reason for the error.

gavinhoward:
I will give it a go. It might take me a few days to get an idea of whether this causes any problems, though.

jonaski:
openSUSE OK: 3.29 with FTS5: https://build.opensuse.org/package/view_file/openSUSE:Factory/sqlite3/sqlite3.spec?expand=1
Fedora ? Cant find the source on https://src.fedoraproject.org/
Megeia OK 3.28 with FTS5: http://sophie.zarb.org/sources/sqlite3/sqlite3.spec/b392aa24d36e83a0e111eb3e54452dcf
Debian OK: 3.29 with FTS5: https://sources.debian.org/src/sqlite3/3.29.0-1/debian/rules/#L62
Ubuntu OK 3.27.2 with FTS5: https://packages.ubuntu.com/disco/sqlite3

Debian jessie wont work (sqlite3 3.8.7), but stretch or above will work.
Ubuntu Xenial will work (sqlite3 3.11.0) and higher.

Looks good.

jonaski:
Switched the centos/appimage worker to compile it's own sqlite https://github.com/jonaski/strawberry-buildbot/commit/c1d480f48d25d9f31acf04d53aebc58c7723b179

gavinhoward:
I haven't found any sql errors after doing several different kinds of searches.

gavinhoward:
Cool! Will pull and start using.

renn0xtek9:
Is there any way I can retrieve the content of a Playlist stored in the strawberry.db using sqlite? Am 30.07.2019 22:52 schrieb "Gavin D. Howard" <notifications@github.com>:Cool! Will pull and start using.

jonaski:
Yes,
```
select * from playlists;
select * from playlist_items;
```

renn0xtek9:
Thanks for your answer,

I did make a python script that does that for clementine some time a go

However the problem was that when I run

|select * from playlist_items; |

|I obtained things like that
|

|113|Library|6554|||||-1||0||||-1|-1|-1.0|-1|||0|-1|-1||||||0|||0|0|-1|-1|0|0|0|0|0|0||||||-1|-1
|

|I assumed from this that 113 is the index of the playlist and 6554 the 
index of the song, which I assume can query from
|

|select * from playlists
|

|and
|

|select * from songs
|

||

|However if I do that, the content of the playlist won't match what I 
have in Clementine. Somehow the id of one table do not correspond to 
those of the other table I will try to switch my whole music collection 
to strawberry, but I am not confident that it will work. If you see any 
mistake I am making / any explanation, don't hesitate to let me know |

renn0xtek9:
I just tried and I have the same problem with strawberry.

If you can give me hint on how to solve this that would be good

jonaski:
You need to request by rowid.
So to get the rowid for playlists so you can use to request the playlist items:
`select rowid, * from playlists;`
The rowid of the playlist is then in the first column. So to get playlist items by the playlist row id 1:
`select rowid, * from playlist_items where playlist = 1;`
The rowid of the song will then be in the first column
So to get the song from the song rowid if the rowid of the song is 1:
`select * from songs where rowid = 1;`

The table columns are not identical to clementine, look here:
https://github.com/jonaski/strawberry/blob/master/data/schema/schema.sql
But you should never rely on the order of the columns as listed there, as they might change in different versions. (Replace * in the previous queries with the columns you actually need from the tables, separated by comma, in the order you want.

You should instead request what you want in the order you want, like this:
`sqlite> select artist, album, title, url from songs where rowid = 19;
Pink Floyd|Wish You Were Here|Shine On You Crazy Diamond - Part One|file:///home/jonas/Test-Music/Test/Wish_You_Were_Here/01_-_Pink_Floyd_-_Wish_You_Were_Here_-_Shine_On_You_Crazy_Diamond_-_Part_One.flac
`
So it returns the data for song rowid 19 in the order I wanted it.

If you need more sql queries, you can look at the code here: https://github.com/jonaski/strawberry/blob/master/src/collection/collectionbackend.cpp (thats for the songs), and https://github.com/jonaski/strawberry/blob/master/src/playlist/playlistbackend.cpp for the playlists. That's where strawberry does all the relevant sql queries, you'll probably understand the raw queries that are used internally whether you know C++/Qt or not. You can also request several songs in one query, look at the CollectionBackend::GetSongsById function.

jonaski:
My initial response was a bit incorrect, fixed now.

PR #237, opened 2019-08-07
Enable About Strawberry and QT menu on macOS

llucps:
I wonder the reason why these menus were disabled.. I thought to enable them.


jonaski:
I don't know why it was disabled, but if it works fine with it, just remove it, all 3 lines

PR #246, opened 2019-08-22
Brazilian Portuguese Language

pepaulo:
This file have a partial brazilian language translate


jonaski:
We use a web service for translations, see https://github.com/jonaski/strawberry/wiki/How-to-translate
You need to modify the existing translations there.
We already have Portuguese translations, but incomplete.
I only include translations that are complete, and so far that's only 2 languages.
I could import your po file, but then I need to know that your version is better and more complete as it will override the existing one probably.

PR #278, opened 2019-10-17
subsonic api returns discNumber, not disc

BlackIkeEagle:
As found in the v1.13.0 xsd, the disc is named discNumber in the api,
currently there is a check for disc which will never be found in the
response.

@see http://www.subsonic.org/pages/inc/api/schema/subsonic-rest-api-1.13.0.xsd

Signed-off-by: BlackEagle <ike.devolder@gmail.com>


jonaski:
Thanks

PR #398, opened 2020-04-06
Fix 'Show in file browser' not working on Linux

plonibarploni:
Ubuntu MATE 19.10 with Caja, clicking 'Show in file browser' results in a Caja showing an error `'/path/to/file.mp3' is not a directory`.
Replacing the current code with Clementine's one-liner works for me.

### DIFF
--- a/src/core/utilities.cpp
+++ b/src/core/utilities.cpp
@@ -331,63 +331,6 @@ QString ColorToRgba(const QColor &c) {
 
 }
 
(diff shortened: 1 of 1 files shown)


jonaski:
The code works with nautilus and dolphin so I'm not removing it.
If Caja doesn't handle a file as a parameter you should fix it so it is passed a directory instead.

PR #399, opened 2020-04-06
Fix 'Show in file browser' for Caja

plonibarploni:
This strips out the filename for Caja, which won't handle a file as a parameter.


jonaski:
Thanks

PR #471, opened 2020-06-22
Update it.po

albanobattistella:


jonaski:
Translations needs to go through Zanata, see:
https://github.com/strawberrymusicplayer/strawberry/wiki/How-to-translate

PR #490, opened 2020-07-24
Simple addition of playlist actions to icon

bongochong:
This is a very simple addition of playlist/playback actions to Strawberry's icon (via the .desktop file). This re-introduces some nice little polish that was present in Clementine's .desktop file. It works, and it allows one to control Strawberry when its main window is minimized, without having to enable the tray icon.


bongochong:
This is an absolutely fantastic project by the way. All it took was trying out Strawberry for five minutes and I completely switched over from Clementine.

Issue #544, opened 2020-09-15
Support for Subsonic API's scrobble method


spezifisch:
Subsonic users can configure their last.fm account in Subsonic's settings and let the Subsonic server do the scrobbling for them.
Issue #416 is related, but what i mean here is another case where you wouldn't have to configure your last.fm account in Strawberry, but the Subsonic server handles communication with last.fm.

Subsonic API has the method `scrobble` for this, see http://www.subsonic.org/pages/api.jsp#scrobble.

To support this Strawberry could:
* When starting a song with `song.Source() == Song::Source::Source_Subsonic`: call `http://subsonic-server/rest/scrobble` with `submission=false` (indicating NowPlaying) and the subsonic song id.
* After the submit delay do a call with `submission=true` and optionally `time=<unix timestamp of song start>`

The Subsonic server then scrobbles this song for you to last.fm.

I really want this feature, so I implemented it at https://github.com/spezifisch/strawberry/commit/a79cb04870d95a121c907ec144a19a944e19b0da. Maybe we could merge this if you consider this feature desirable.

My changes are:
* add `SubsonicService::Scrobble` method to make the request to the Subsonic server
* add `SubsonicScrobbler` along the other scrobblers

The following improvements are on my todo list:
* ~~send `time` parameter of song start~~
* ~~includes need to be cleaned up (i just copied listenbrainzscrobbler as a base for my changes)~~
* test with Airsonic and Subsonic. I'm using [gonic](https://github.com/sentriz/gonic) as my Subsonic server, which also supports the method `scrobble` and tested only with it.

I'd appreciate any input on how my patch could be improved.

jonaski:
I'm fine with this but you need to create a subclass of SubsonicBaseRequest called SubsonicScrobbleRequest so you can free the QNetworkReply and handle errors.
I think it also should be a checkbox in the subsonic settings to enable it.

spezifisch:
I added PR #545 so we can discuss my code better and so I can use the CI.

> I'm fine with this but you need to create a subclass of SubsonicBaseRequest called SubsonicScrobbleRequest so you can free the QNetworkReply and handle errors.

I've added SubsonicScrobbleRequest.
I also added the queueing code back in to avoid hanging scrobbling requests due to shaky internet from blocking further submissions before they time out.

[...]

PR #545, opened 2020-09-17
Add Subsonic scrobble support

spezifisch:
This is a work in progress for #544.

This PR adds:
* SubsonicScrobbler class along the other scrobblers
* SubsonicService::Scrobble method to be called from SubsonicScrobbler
* SubsonicScrobbleRequest class to make the requests to the Subsonic server

### DIFF
--- a/src/scrobbler/subsonicscrobbler.cpp
+++ b/src/scrobbler/subsonicscrobbler.cpp
@@ -33,6 +33,7 @@
 #include "core/logging.h"
 #include "core/closure.h"
 #include "internet/internetservices.h"
+#include "settings/subsonicsettingspage.h"
 #include "subsonic/subsonicservice.h"
 
 #include "audioscrobbler.h"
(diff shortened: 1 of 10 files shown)


src/scrobbler/subsonicscrobbler.cpp
```diff
@@ -0,0 +1,111 @@
...
+  submitted_(false) {
+  ReloadSettings();
+}
+
+SubsonicScrobbler::~SubsonicScrobbler() {
+
+}
+
+void SubsonicScrobbler::ReloadSettings() {
+
```
jonaski, line 56: You can read serversidescrobbling from the subsonic settings here and set enabled_ based on that so that it can be used by AudioScrobbler.

src/subsonic/subsonicservice.cpp
```diff
@@ -379,6 +383,25 @@ void SubsonicService::CheckConfiguration() {
 
 }
 
+void SubsonicService::Scrobble(QString song_id, bool submission, QDateTime time) {
+
+  if (!server_side_scrobbling()) {
+      return;
```
jonaski, line 389: Indent 2 spaces

jonaski:
Looks good. I tested with Airsonic and it works.

jonaski (changes requested):

jonaski (approved):

jonaski:
Thanks! I fixed reloading the subsonic scrobbler settings.

Issue #577, opened 2020-11-02
Song progress on tray icon


yavuzmert:
The only feature left that I personally miss from clementine is track progress on system tray icon. I see that most of the code is already there but commented out and grey icons were missing. I created grey icons from original strawberry icons.
I changed the direction to top-right to bottom left to better match the strawberry icon. (Going through the bottom of the strawberry). The result is acceptable I guess.
If you don't like the feature and intentionally commented it out, maybe we can make it optional.
If you approve the changes I will make a PR for this.
Thanks.
PS: I changed the mac source also but I cannot test that.
**Screenshots:**
![1](https://user-images.githubusercontent.com/19837772/97929591-92bce280-1d7a-11eb-95cf-cc4950ac55aa.png)
![2](https://user-images.githubusercontent.com/19837772/97929595-95b7d300-1d7a-11eb-869f-4756b5b56bac.png)
![3](https://user-images.githubusercontent.com/19837772/97929602-97819680-1d7a-11eb-9646-887aa1ff42b6.png)

jonaski:
An option in the behavior settings is fine, it's correct that I just commented it out because I didn't think it looked good with the strawberry icon, and never bothered to deal with it.

jonaski:
I can test it on mac if you make a PR.

jonaski:
Thanks!

PR #578, opened 2020-11-03
Re-enable progress on system tray icon

yavuzmert:
#577


src/core/qtsystemtrayicon.cpp
```diff
@@ -54,7 +54,11 @@ QtSystemTrayIcon::QtSystemTrayIcon(QObject *parent)
       action_mute_(nullptr) {
 
   app_name_[0] = app_name_[0].toUpper();
-
+  QIcon theme_icon_grey =
```
jonaski, line 57: One line please

src/core/qtsystemtrayicon.cpp
```diff
@@ -54,7 +54,11 @@ QtSystemTrayIcon::QtSystemTrayIcon(QObject *parent)
...
+      IconLoader::Load("strawberry-grey");
+  if (! theme_icon_grey.isNull()) {
```
jonaski, line 59: remove space after !

jonaski (changes requested):
It would be nice if you could make a setting for this in the top options of the behavior settings (settings/appearancesettingspage)

yavuzmert:
Hi Jonas, 
Setting and icon progress seem to work - at least on my system :)
I put the checkbox at the bottom of appearance settings, later i saw that you said top of the page. I think bottom is alright but you may reorganize it and put wherever you see appropriate.

src/core/systemtrayicon.cpp
```diff
@@ -47,31 +49,36 @@ SystemTrayIcon::SystemTrayIcon(QObject *parent)
...
-
   QRect rect(icon.rect());
 
-  // The angle of the line that's used to cover the icon.
-  // Centered on rect.topRight()
-  double angle = double(100 - song_progress()) / 100.0 * M_PI_2 + M_PI;
-  double length = sqrt(pow(rect.width(), 2.0) + pow(rect.height(), 2.0));
+  QSettings s;
+  s.beginGroup(AppearanceSettingsPage::kSettingsGroup);
+  bool trayicon_progress = s.value(AppearanceSettingsPage::kTrayIconProgress, false).toBool();
```
jonaski, line 56: Move this to MainWindow::ReloadSettings() so we don't reload the setting each time the icon is painted.
Then add a public method in SystemTrayIcon to call when the setting is reloaded.

src/settings/appearancesettingspage.cpp
```diff
@@ -197,6 +199,10 @@ void AppearanceSettingsPage::Load() {
   ui_->blur_slider->setValue(s.value(kBlurRadius, kDefaultBlurRadius).toInt());
   ui_->opacity_slider->setValue(s.value(kOpacityLevel, kDefaultOpacityLevel).toInt());
 
+#if !defined(Q_OS_WIN)
```
jonaski, line 202: Windows uses QSystemTrayIcon  too so this isn't needed.

jonaski (changes requested):

jonaski:
Sorry I meant behavior settings, settings/behavioursettingspage. I think it makes more sense since that's where the system tray icon options are.

yavuzmert:
Alright I moved it to behavioursettingspage, I set it enabled when tray icon is checked. 
But only problem I see is, on MacOs, tray icon settings are hidden.
We can show its checkbox also on MacOs although "Show system tray icon" checkbox is hidden, but then we must set its label which makes sense all platforms (maybe "show progress on icon" instead of "show progress on system tray icon")

jonaski:
The track progress doesn't work on macOS so you can just exclude that option. It's not really a system tray icon on mac at all, it's just the context menu on the dock.

src/core/mainwindow.cpp
```diff
@@ -1032,8 +1032,11 @@ void MainWindow::ReloadSettings() {
 
   s.beginGroup(AppearanceSettingsPage::kSettingsGroup);
   int iconsize = s.value(AppearanceSettingsPage::kIconSizePlayControlButtons, 32).toInt();
+  bool trayicon_progress = s.value(AppearanceSettingsPage::kTrayIconProgress, false).toBool();
```
jonaski, line 1035: Move to behavioursettings

jonaski (changes requested):

jonaski (approved):

Issue #515, opened 2020-08-19
Currently playing song text is blurry on high DPI display


eggjoystick:
For general technical questions and help with technical issues please use the forum on https://forum.strawberrymusicplayer.org/

**Describe the bug**
Text for the currently playing song (with pulsating background) is blurry and pixelated on a 3840x2160 display with 2x scaling.

**To Reproduce**
Use strawberry on a high DPI display and Cinnamon desktop environment with user interface scaling set to Double (Hi-DPI). Play a track and notice the blurry text for the current song.

**Expected behavior**
Text should remain crisp as it does for all text

**Screenshots:**
![strawberry-blurry-text](https://user-images.githubusercontent.com/58571795/90688232-c5ebdd00-e233-11ea-9602-c36ea525efea.png)

**System Information:**
 - Operating system: MX Linux 19.2 64-bit
 - Strawberry Version: 0.7.2
 - Kernel: 5.4.0-3-amd64
 - Cinnamon 4.2.4
 - Resolution: 3840x2160~60Hz

**Additional context**
All text not pertaining to currently playing song remains crisp

jonaski:
I don't have a hiDPI screen to test on here, but you can try the following:
Does disabling "show glowing animation" in the playlist settings change anything?
Do you know how to compile from source?
Remove these 3 lines from playlistview.cpp:
1. `style_(new PlaylistProxyStyle())` on https://github.com/strawberrymusicplayer/strawberry/blob/master/src/playlist/playlistview.cpp#L135
2. `setStyle(style_);` on https://github.com/strawberrymusicplayer/strawberry/blob/master/src/playlist/playlistview.cpp#L179
3. `delete style_;` on https://github.com/strawberrymusicplayer/strawberry/blob/master/src/playlist/playlistview.cpp#L215
Does that fix it?

eggjoystick:
Hi, thanks for replying so quickly.

Disabling the "show glowing animation" playlist setting does not resolve the blurry text.

I've managed to delete the 3 lines from playlistview.cpp but I could use some instructions on how to compile from source. I am comfortable using the command line.

jonaski:
https://github.com/strawberrymusicplayer/strawberry/wiki/Compiling-on-Debian

eggjoystick:
Thanks for the instructions. I was able to follow them, compile from source, and run strawberry. Unfortunately, the blurry text remains.

jonaski:
[...]

PR #587, opened 2020-11-10
Fix HiDPI scaling for glow animation and drag over playlist

fbugno:
This set the proper scaling and pixel ratio of QPixmap widgets
used as cached objects, in a effort to solve this following issue:

https://github.com/strawberrymusicplayer/strawberry/issues/515

Most of cached objects uses a custom QPaint instead of the default
painter object from the parent widget. The problem is that, unlike
the painter from the parent object, set by the main application,
and that has DPI and scaling settings from the device, these custom
QPainters don't know about the underlying device, thus uses a
scale of 1 to render artifacts.

When a cached object "edited" by a custom QPaint along his pipeline
where used on a paint or drawrow routine, his stored image is distorted
and burred in a effort to resize it to the display configuration.

### DIFF
--- a/src/playlist/playlistview.cpp
+++ b/src/playlist/playlistview.cpp
@@ -215,7 +215,7 @@ PlaylistView::PlaylistView(QWidget *parent)
   dynamic_controls_->hide();
 
   // To proper scale all pixmaps
-  device_pixel_ratio = this->devicePixelRatioF();
+    device_pixel_ratio_ = devicePixelRatioF();
 
   // For fading
   connect(fade_animation_, SIGNAL(valueChanged(qreal)), SLOT(FadePreviousBackgroundImage(qreal)));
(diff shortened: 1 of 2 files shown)


src/playlist/playlistview.h
```diff
@@ -220,6 +220,7 @@ class PlaylistView : public QTreeView {
   Playlist *playlist_;
   PlaylistHeader *header_;
 
+  qreal device_pixel_ratio;
```
jonaski, line 223: device_pixel_ratio_

src/playlist/playlistview.cpp
```diff
@@ -214,6 +214,9 @@ PlaylistView::PlaylistView(QWidget *parent)
 
   dynamic_controls_->hide();
 
+  // To proper scale all pixmaps
+  device_pixel_ratio = this->devicePixelRatioF();
```
jonaski, line 218: Can we omit this-> here? Can we call devicePixelRatioF() directly on the functions?

jonaski (changes requested):

jonaski:
Thanks for this fix!

src/playlist/playlistview.cpp
fbugno: Sure, I will do a commit to reflect that and the variable above.

jonaski (approved):

PR #592, opened 2020-11-18
Finished the Swedish translation

jony0008:
I have finished the Swedish translation of Strawberry but I could not upload it to Zanata, I hope this will be more successful. Upload and overwrite the .po file in Zanata for future changes.


jony0008:
The file has already been added. I will close this pl.

PR #613, opened 2020-12-06
[Cover manager] Open cover when double clicking

### LINKED ISSUE #612: [Cover manager] Double clicking a cover adds the album to the playlist
Using 0.8.4 from PPA on Linux x64

When opening the cover manager, if you double click on a cover, the album is added to the playlist.  
Instead, the cover should be displayed at real size.

Cqoicebordel:
This is just a proposed change to fix #612.  
It has not been tested, and it doesn't handle the function `AlbumCoverManager::AlbumDoubleClicked()`. Maybe changing the content of this function would be better, but I'll let you judge that :)

### DIFF
--- a/src/covermanager/albumcovermanager.cpp
+++ b/src/covermanager/albumcovermanager.cpp
@@ -204,7 +204,7 @@ void AlbumCoverManager::Init() {
   connect(ui_->export_covers, SIGNAL(clicked()), SLOT(ExportCovers()));
   connect(cover_fetcher_, SIGNAL(AlbumCoverFetched(quint64, QUrl, QImage, CoverSearchStatistics)), SLOT(AlbumCoverFetched(quint64, QUrl, QImage, CoverSearchStatistics)));
   connect(ui_->action_fetch, SIGNAL(triggered()), SLOT(FetchSingleCover()));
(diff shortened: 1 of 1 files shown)


jonaski:
I'm fine with the intended change. 
However your change won't work as GetSingleSelectionAsSong() is based on context_menu_items_which is set when you right click based on the albums selected.
So you need to modify the AlbumDoubleClicked() function to get the song that belong to the QModelIndex passed to that function.
Then you can call AlbumCoverChoiceController::ShowCover with that song.

Cqoicebordel:
I followed your instruction, but as I said in the commit message, I'm not sure of my code, at all.

jonaski:
This works, but will query SQL again which isn't necessary since the model already holds the required data.

To avoid querying SQL:

```
QListWidgetItem *item = static_cast<QListWidgetItem*>(index.internalPointer());
if (!item) return;
album_cover_choice_controller_->ShowCover(ItemAsSong(item));

```

Cqoicebordel:
Ok, I wouldn't have found it :D  
Thanks for explaining though :)

I'm closing this PR, since it's 100% your code :)

Issue #624, opened 2021-01-03
Last.fm last played date import is incorrect


esalgado:
Hello! 

Sorry for my late reply... I have finally compiled using QT6, and the import last playcount made it to the end :+1: 

But I think that instead of this being the last played date, it is the first played date:
![Screenshot_20210102_122414](https://user-images.githubusercontent.com/977612/103456357-2fedb600-4ced-11eb-9d65-2be173145e78.png)

Can it be that the function is continously updating the play date for a particular song, in the order that the Last.fm API sends them? All those 51k plays, is all my play history since 2005, and maybe it is being sent by Last.Fm in reverse chronological order, or maybe random?

I don't know if I have to open a bug for this?

Thanks for the app, I love it :)

_Originally posted by @esalgado in https://github.com/strawberrymusicplayer/strawberry/issues/247#issuecomment-753462788_

esalgado:
#625 This fixes it!

jonaski:
The `user.getRecentTracks` API method returns all scrobble history instead of being unique for each song, there isn't a better API method to use, so maybe we should just remove the last played import, and just keep the play count, it's too much to retrieve the entire scrobble history.

millette:
@jonaski Thanks for the tip, I'm an old-time last.fm user, new strawberry user and importing last player from last.fm doesn't work for me, too long a history I guess. Requesting only play counts seems to work, great!

I'd still like to import last played, even if it's not complete. Should I open another issue about my bug?

PR #625, opened 2021-01-03
Last.fm last played date import: change date if it is newer than the …

esalgado:
…one in database

In response to this issue: https://github.com/strawberrymusicplayer/strawberry/issues/624#issue-777738164

I have modified the code and tested, and it is working :+1: 

Thanks,

### DIFF
--- a/src/collection/collectionbackend.cpp
+++ b/src/collection/collectionbackend.cpp
@@ -1380,24 +1380,15 @@ void CollectionBackend::UpdateLastPlayed(const QString &artist, const QString &a
   QSqlDatabase db(db_->Connect());
 
   for (const Song &song : songs) {
-    QSqlQuery q(db);
-    q.prepare(QString("SELECT lastplayed FROM %1 WHERE ROWID = :id").arg(songs_table_));
-    q.bindValue(":id", song.id());
-    q.exec();
-    if (db_->CheckErrors(q)) continue;
-    while (q.next()) {
-        if (q.value(0).toInt() >= lastplayed){
-            //qLog(Debug) << "Database data is more recent than submited one for" << artist << album << title;
-            return;
-        }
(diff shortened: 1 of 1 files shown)


esalgado:
Last played date for the first songs with date coincides with the time I started using Last.fm:

last played

src/collection/collectionbackend.cpp
```diff
@@ -1381,11 +1381,23 @@ void CollectionBackend::UpdateLastPlayed(const QString &artist, const QString &a
 
   for (const Song &song : songs) {
     QSqlQuery q(db);
+    q.prepare(QString("SELECT lastplayed FROM %1 WHERE ROWID = :id").arg(songs_table_));
```
jonaski, line 1384: We already got this in GetSongsBy() where Song::InitFromQuery is used. Just compare with song.lastplayed()

jonaski (changes requested):

src/collection/collectionbackend.cpp
esalgado: Like this?

jonaski: Yes but continue instead of return, and remove the commented debug.
The code style should be consistent with the rest, space between ) and {. We also use 2 spaces for indent, not 4.

esalgado: OK, I will do it. If I change return to continue, I need to invert the comparison or negate it with ! , right?

Sorry, it is my first time with QT and C++!

jonaski: No. With continue we skip the updating SQL part and continue to the next song, ie.: another song on a different album.

jonaski: https://www.w3schools.com/cpp/cpp_break.asp

jonaski (approved):

PR #626, opened 2021-01-04
CMakeLists.txt: Make gstreamer/vlc backend search optional

Polynomial-C:
this doesn't change the default of searching both backend requirements but gives packagers the opportunity to selectively disable search of one of the backends and emits a fatal error if search for both backends have been disabled.

Gentoo-bug: https://bugs.gentoo.org/757780
Signed-off-by: Lars Wendler <polynomial-c@gentoo.org>


jonaski:
The pkg_check_modules for gstreamer and vlc does not have REQUIRED, so cmake won't fail if they are not found.
If VLC and gstreamer are found, they can still be specifically be disabled with existing options:
`-DENABLE_VLC=OFF` for VLC and -`DENABLE_GSTREAMER=OFF` for gstreamer, then the libraries won't be linked.

Polynomial-C:
Oh, right. Didn't see these options and they serve exactly that purpose I wanted to have. Sorry for the useless noise...

PR #637, opened 2021-01-24
Connection syntax migration

jonaski:


gavinhoward:
Hi! I won't be able to do the review today, but I will do it as soon as I can, probably tomorrow.

jonaski:
I did this ages ago, but I was hesitant to change to the new syntax then.
Even though the old connection syntax is still available in Qt 6, I think they plan to deprecate it in the future, I saw this: https://bugreports.qt.io/browse/QTBUG-76375
It's also much safer with the new syntax since signal-slot connections are checked at compile-time. So might as well change it now.

Can merge it when we've done some more testing. The most critical thing to test is probably tag-editing, both the tag editor and the playlist double click and right click edits.

- There are only a couple of uses left, one is in `src/globalshortcuts/globalshortcuts.cpp` where a helper function is used, wasn't sure how to replace the `SIGNAL()` in the function parameter to keep the helper function. And one is in `src/core/mainwindow.cpp` for the playback resume feature, this is one-time connection only from player using `NewClosure`'s. Will probably deal with that later as I want to improve the resume playback feature anyway.
- `NewClosure` is only required for `QFuture` direct connections for dealing with `QFutureWatcher` now, the rest is replaced with lambdas. Most of this was done previously, but I've replaced a few `TagReply` related ones now in this PR. Can consider removing the rest of the closure methods when the closure unit tests are updated to test `QFuture`.
- I had to rename a few slots that were incompatible with the signals.
- I had to move some slots to public because they were incorrectly placed in private.
- Some slots that had incompatible optional parameters could not be used with the new synax, so I simply made helper functions (can use lambda too).
- I removed const and const references from signals, my understand is that this isn't needed for signals.
- A few code style and header guard related changes too.

jonaski:
Reverting back to `NewClosure` for `TagReaderReply` as it caused issues with lambdas.

gavinhoward:
Just so I have the right context, this is about a syntax switch between Qt 5 and Qt 6? I will need to do some research about that.

jonaski:
No, it's old Qt 4 syntax vs. new Qt 5/6 syntax:
https://wiki.qt.io/New_Signal_Slot_Syntax

jonaski:
The rest seem to be working as far as I have tested, so I'm just merging it. Appreciate if you can have a look at it anyway when you have time. Particularity the changes I did in `edittagdialog.cpp` and and `lineedit.cpp`, and possible do some testing too.

gavinhoward:
I will do. Sorry that I am taking so long to remember how to work with Qt.

gavinhoward:
I had a segfault with the following logs.

```
19:49:42.782 ERROR MusixmatchLyricsProvider:221     MusixmatchLyrics: "Error transferring https://www.musixmatch.com/lyrics/trevor-rabin/ben-finds-pipe - server replied: Not Found (203)"
19:50:48.735 WARN  unknown                          libpng warning: iCCP: known incorrect sRGB profile 
19:50:50.490 ERROR MusixmatchLyricsProvider:221     MusixmatchLyrics: "Json lyrics is missing body."
19:57:58.486 ERROR MusixmatchLyricsProvider:221     MusixmatchLyrics: "Error transferring https://www.musixmatch.com/lyrics/jonathan-thulin/jekyll-and-hyde - server replied: Not Found (203)"
19:58:05.210 WARN  unknown                          QObject::startTimer: Timers can only be used with threads started with QThread 
Segmentation fault (core dumped)
```

gavinhoward:
I found another one, and the common theme is the `QObject::startTimer: Timers can only be used with threads started with QThread` message. I am starting to look into it.

jonaski:
That specific error can come from a `QTimer::singleShot()`.
Are you testing the latest master with commit https://github.com/strawberrymusicplayer/strawberry/commit/819800d3a415ceb5898f640af6e2715e89bca44b
I found that problem with the moodbar and reverted the lambda conversions back to NewClosure, and I can't reproduce that error anymore.

jonaski:
Should be able to find all the lambda connects with: `grep -r '.*connect(.*].*{'`

jonaski:
I understand the problem now with the lambda connects I did for the TagReaderReply and the moodbar. The callback function was called from the thread of the signal object instead of the main thread.
Need to do:
`connect(threadobject, &ThreadObject::Finished, this, [this]() { TestFinished(); });`
instead of:
`connect(threadobject, &ThreadObject::Finished, [this]() { TestFinished(); });`

gavinhoward:
Ah, okay. Sorry. I had to go to bed before I could trace the problem. I was already running strawberry in `gdb` to see if I could reproduce the problem with a stack trace.

gavinhoward:
I've been running `strawberry` for a while with the update, and no problem so far.

Issue #296, opened 2019-11-10
Spectrum-based fingerprinting as the main track identification in database and playlists


v-fox:
### Is your feature request related to a problem? Please describe.

The biggest hassle in a music collection and in the clementine in particular is tethering to file paths and names: rename or move one directory and mess up all your playlists containing any of affected files.

### Describe the solution you'd like

This can be completely solved by going semantic: using spectral meta-data as fingerprinting stamps instead of names, tags and encoding formats. In particular, the glorious moodbar can double for that. IF it could guarantee reproducible stamps for the same file under all circumstances.

### Describe alternatives you've considered

There was [a Clementine pull request](https://github.com/clementine-player/Clementine/pull/5135) for limited type of fingerprinting but it does not account for whole content and only tracks renames of files under unchanged directory structures.

### Additional context

This would also allow for copying files at will while preserving original playlist ordering. Adding a tag as secondary identifier may allow tracking "relations" between files such as down-converted transcodings for mobile devices (like FLAC->AAC/OPUS) inheriting ID of their original in a tag, so they could be put in the same spot in a playlist on a different device. That would immensely help with sharing playlists as new age gigantic high-quality alternatives to "mix tapes" of the past.

Especially if a compatible Android player would manifest or at least a script for generating legacy playlists out of that meta-data after "settling" files in whatever silly directory structure mobiles have (like being unable to use ':' in file-paths or forcibly lumping gigabytes of tracks together by artist/title in an incomprehensible mess).

v-fox:
I probably should have elaborated more on the proposed implementation:
* Instead of filename use unique hash (with diverse enough algorithm, ideally, one that could be changed in future, so having "algorithm ID" or "protocol version" somewhere would be nice) of moodbar "file" (not actual file, a "binary tag" in the database that is cached as a file and/or as tag in supported formats). Moodbar-generation algorithm also could be changed in the future that would change hashes, so particular pair of moodbar function and hashing algorithm would constitute a "protocol version".
    * In addition to moodbar as primary defining characteristic hashed data may also include other defining informations, such as format (PCM, DSD, MIDI, etc.) and format-specific parameters (bit-depth, sampling rate, exact MIDI type, etc.), but NOT any changeable information, such as other meta-data (name, tags, filesystem properties, etc.). Although it would be nice to know if 2 files result in bit-perfect or near-perfect spectrum recreation.
[...]

PR #721, opened 2021-06-01
Add song fingerprinting and tracking

jonaski:
This PR adds song fingerprinting and tracking and improves the collection update progress.

@gavinhoward If you have time to look at this it would be nice.

I will probably merge this over the weekend.

**The collection update progress:**

It now shows the progress based on files instead of just directories, and calculates the number of files before starting the scanning, this fixes the old behavior where it was annoyingly going up and down while scanning because of new found directories.
This makes the collection update a bit slower while calculating the number of files, but not much, it will show 0% while it is calculating the number of files.
For each subdir it passes to ScanSubdirectory, it also passes the number of files/directories within that subdir excluding the symlinks and the nomedia files. It only really needs this for AddToProgress in case the subdir is skipped based on the mtime, this could probably be changed to do this check directly when calculating files.

**Song fingerprinting and tracking:**

- In the collection settings there is a new setting to enable song fingerprinting and tracking, if it is enabled, the mark deleted songs as unavailable is always on since songs tracking is useless without it. The song fingerprinting is off by default mainly because it will make a full scan much slower and some users might complain.
- There are now 2 new fields in the DB: "fingerprinting" and "lastseen". The lastseen field is used to expire songs that are marked unavailable so the DB does not get filled up with unavailable songs.
- There is a new periodic timer in the collection watcher that triggers an incremental scan if it has not run for a long time, this is used to update the lastseen field, and also expire unavailable songs that are over a certain number of days that is configurable in the collection settings.
- If fingerprinting is enabled and one or more songs in the DB is missing a fingerprints, it forces a full scan so it can calculate the fingerprints and update the songs in the DB with the fingerprint.
- Songs with empty fingerprint is cached for a transaction so that it doesn't have to query the DB for each subdir.
- An incremental scan will only recalculate the fingerprint if the mtime is changed.
- The main changes to the collection watcher is in ScanSubdirectory():
1. When iterating the files, instead of accepting all files that are not hidden, it does an extra IsMediaFileBlocking() and skips the file if it's not a audio file. This is mainly done here because the fingerprinting needs to be done before ScanNewFile(), and this prevents trying to calculate a fingerprint for a non-audio file, I can't think of any reason to include them in files_on_disk, since image's are still found in the first loop, and cue's are calculated based on the audio filename. TagLib does the extension based check first, so if the directories contain mostly audio files, it shouldn't take much longer to do this extra check here.
2. When iterating files_on_disk, it first does the normal FindSongByPath() to see if the file exist by path, the change here is to include the fingerprint if it is either missing, or the mtime is changed it will recalculate the fingerprint too.
3. If the file is not found in the DB by path it calculates fingerprint and calls FindSongByFingerprint(), this function queries the DB for the songs with the same fingerprint independent of collection directories, if any songs is found in the DB with the same fingerprint, it only uses it if path is the same (this will normally not be the case here since we already queries by path), or the path is different but the file currently in the DB does not exist. This allows duplicate audio files in different directories, i.e: different albums with identical songs. It adds the 
4. If the song is not found by fingerprint, it adds the file to the DB as a new song like before but includes the fingerprint it calculated.
5. I added a new function called UpdateSong(), this function displays better debugging information about why the song was updated.
6. There is some code duplication in ScanSubdirectory, the FindSongByPath() and the FindSongByFingerprint if conditions share some of the same code, this could possibly be improved later.
7. In the case where it for some reason fails to calculate the fingerprint for a song, it sets the fingerprint to NONE so it does not cause the fingerprint to be recalculated on a scan.

Fixes https://github.com/strawberrymusicplayer/strawberry/issues/296


gavinhoward:
I successfully built the branch on my system and did a full collection scan.

The progress is *so* much better. I love it.

While I do not use Strawberry's playlist functionality (I build mine by hand and save them to text files), I'll give it a whirl in an attempt to test this.

First pass at review looks good, but I will dive deeper.

src/core/song.cpp
```diff
@@ -1519,9 +1544,12 @@ bool Song::IsMetadataEqual(const Song &other) const {
...
 }
 
-bool Song::IsMetadataAndArtEqual(const Song &other) const {
+bool Song::IsMetadataAndMoreEqual(const Song &other) const {
 
-  return IsMetadataEqual(other) && d->art_automatic_ == other.d->art_automatic_ && d->art_manual_ == other.d->art_manual_;
+  return IsMetadataEqual(other) &&
+         d->fingerprint_ == other.d->fingerprint_ &&
+         d->art_automatic_ == other.d->art_automatic_ &&
+         d->art_manual_ == other.d->art_manual_;
```
gavinhoward, line 1552: 👍

src/core/song.h
```diff
@@ -231,9 +231,12 @@ class Song {
   qint64 mtime() const;
   qint64 ctime() const;
 
+  QString fingerprint() const;
+
   int playcount() const;
   int skipcount() const;
   qint64 lastplayed() const;
+  qint64 lastseen() const;
```
gavinhoward, line 239: 👍

src/core/song.h
```diff
@@ -345,9 +348,12 @@ class Song {
   void set_ctime(qint64 v);
   void set_unavailable(bool v);
 
+  void set_fingerprint(const QString &v);
+
   void set_playcount(int v);
   void set_skipcount(int v);
   void set_lastplayed(qint64 v);
+  void set_lastseen(qint64 v);
```
gavinhoward, line 356: 👍

src/core/song.h
```diff
@@ -365,7 +371,7 @@ class Song {
 
   // Comparison functions
   bool IsMetadataEqual(const Song &other) const;
-  bool IsMetadataAndArtEqual(const Song &other) const;
+  bool IsMetadataAndMoreEqual(const Song &other) const;
```
gavinhoward, line 374: 👍

src/settings/collectionsettingspage.cpp
```diff
@@ -80,6 +80,8 @@ CollectionSettingsPage::CollectionSettingsPage(SettingsDialog *dialog)
   QObject::connect(ui_->add, &QPushButton::clicked, this, &CollectionSettingsPage::Add);
   QObject::connect(ui_->remove, &QPushButton::clicked, this, &CollectionSettingsPage::Remove);
 
+  QObject::connect(ui_->song_tracking, &QCheckBox::toggled, this, &CollectionSettingsPage::SongTrackingToggled);
+
```
gavinhoward, line 84: 👍

src/settings/collectionsettingspage.cpp
```diff
@@ -124,6 +126,15 @@ void CollectionSettingsPage::CurrentRowChanged(const QModelIndex &idx) {
...
 
+void CollectionSettingsPage::SongTrackingToggled() {
+
+  ui_->mark_songs_unavailable->setEnabled(!ui_->song_tracking->isChecked());
+  if (ui_->song_tracking->isChecked()) {
+    ui_->mark_songs_unavailable->setChecked(true);
+  }
+
+}
+
```
gavinhoward, line 137: 👍

src/settings/collectionsettingspage.cpp
```diff
@@ -157,7 +168,9 @@ void CollectionSettingsPage::Load() {
   ui_->show_dividers->setChecked(s.value("show_dividers", true).toBool());
   ui_->startup_scan->setChecked(s.value("startup_scan", true).toBool());
   ui_->monitor->setChecked(s.value("monitor", true).toBool());
-  ui_->mark_songs_unavailable->setChecked(s.value("mark_songs_unavailable", false).toBool());
+  ui_->song_tracking->setChecked(s.value("song_tracking", false).toBool());
+  ui_->mark_songs_unavailable->setChecked(ui_->song_tracking->isChecked() ? true : s.value("mark_songs_unavailable", true).toBool());
+  ui_->expire_unavailable_songs_days->setValue(s.value("expire_unavailable_songs", 60).toInt());
```
gavinhoward, line 173: 👍

src/settings/collectionsettingspage.cpp
```diff
@@ -216,7 +229,9 @@ void CollectionSettingsPage::Save() {
   s.setValue("show_dividers", ui_->show_dividers->isChecked());
   s.setValue("startup_scan", ui_->startup_scan->isChecked());
   s.setValue("monitor", ui_->monitor->isChecked());
-  s.setValue("mark_songs_unavailable", ui_->mark_songs_unavailable->isChecked());
+  s.setValue("song_tracking", ui_->song_tracking->isChecked());
+  s.setValue("mark_songs_unavailable", ui_->song_tracking->isChecked() ? true : ui_->mark_songs_unavailable->isChecked());
+  s.setValue("expire_unavailable_songs", ui_->expire_unavailable_songs_days->value());
```
gavinhoward, line 234: 👍

src/settings/collectionsettingspage.h
```diff
@@ -76,6 +76,7 @@ class CollectionSettingsPage : public SettingsPage {
   void Remove();
 
   void CurrentRowChanged(const QModelIndex &idx);
+  void SongTrackingToggled();
```
gavinhoward, line 79: 👍

src/settings/collectionsettingspage.ui
```diff
@@ -92,13 +92,87 @@
         </property>
        </widget>
       </item>
+      <item>
+       <widget class="QCheckBox" name="song_tracking">
+        <property name="text">
+         <string>Song fingerprinting and tracking</string>
+        </property>
+       </widget>
+      </item>
```
gavinhoward, line 101: 👍

src/settings/collectionsettingspage.ui
```diff
@@ -92,13 +92,87 @@
...
+            <size>
+             <width>40</width>
+             <height>20</height>
+            </size>
+           </property>
+          </spacer>
+         </item>
+        </layout>
+       </widget>
+      </item>
```
gavinhoward, line 175: 👍

src/collection/collectionbackend.cpp
```diff
@@ -1545,3 +1591,44 @@ void CollectionBackend::UpdateSongRatingAsync(const int id, const double rating)
...
+    QMutexLocker l(db_->Mutex());
+    QSqlDatabase db(db_->Connect());
+
+    QSqlQuery q(db);
+    q.prepare(QString("UPDATE %1 SET lastseen = :lastseen WHERE directory_id = :directory_id AND unavailable = 0").arg(songs_table_));
+    q.bindValue(":lastseen", QDateTime::currentDateTime().toSecsSinceEpoch());
+    q.bindValue(":directory_id", directory_id);
+    q.exec();
+    db_->CheckErrors(q);
+  }
```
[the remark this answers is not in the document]
jonaski: Yes

src/collection/collectionwatcher.cpp
```diff
@@ -73,25 +78,40 @@ CollectionWatcher::CollectionWatcher(Song::Source source, QObject *parent)
...
+  periodic_scan_timer_->setInterval(86400 * kMsecPerSec);
+  periodic_scan_timer_->setSingleShot(false);
+
+  QStringList image_formats = ImageUtils::SupportedImageFormats();
+  for (const QString &format : image_formats) {
+    if (!sValidImages.contains(format)) {
+      sValidImages.append(format);
+    }
+  }
+
```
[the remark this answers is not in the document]
jonaski: `ImageUtils::SupportedImageFormats()` is a helper for supported formats/extensions for images, based on what Qt plugins the user have.

src/collection/collectionwatcher.cpp
```diff
@@ -156,8 +185,10 @@ CollectionWatcher::ScanTransaction::ScanTransaction(CollectionWatcher *watcher,
       incremental_(incremental),
       ignores_mtime_(ignores_mtime),
       mark_songs_unavailable_(mark_songs_unavailable),
+      expire_unavailable_songs_days_(60),
       watcher_(watcher),
       cached_songs_dirty_(true),
+      cached_songs_missing_fingerprint_dirty_(true),
```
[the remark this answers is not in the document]
jonaski: It's a QList so no initialization required.

src/collection/collectionwatcher.cpp
```diff
@@ -352,21 +411,26 @@ void CollectionWatcher::ScanSubdirectory(const QString &path, const Subdirectory
     QString real_path = path_info.symLinkTarget();
     for (const Directory &dir : qAsConst(watched_dirs_)) {
       if (real_path.startsWith(dir.path)) {
-        t->AddToProgress(1);
```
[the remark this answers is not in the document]
jonaski: Yes but not if it's symlinked to a different collection directory, it also skips it in FilesCountForPath(), so it's not part of the total files/folder, this way it only includes the files it actually scans for tags or fingerprint.

src/collection/collectionwatcher.cpp
```diff
@@ -422,15 +491,16 @@ void CollectionWatcher::ScanSubdirectory(const QString &path, const Subdirectory
   QSet<QString> cues_processed;
 
   // Now compare the list from the database with the list of files on disk
-  for (const QString &file : files_on_disk) {
+  QStringList files_on_disk_copy = files_on_disk;
+  for (const QString &file : files_on_disk_copy) {
+
```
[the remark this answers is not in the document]
jonaski: I didn't like the approach where it modifies the same list it was iterating, not sure if that's safe: `files_on_disk.removeAll(file);`.

gavinhoward (commented):
Looks good, but I think you missed one default value change for `lastseen` in `data/schema/schema-14.sql`.

gavinhoward (approved):

jonaski:
Bugfixes:
- Only emit UpdateLastSeen() on full or incremental scan (incremental_ or ignores_mtime_ is set), not when RescanPathsNow() is triggered to check only specific paths. , so it only updates lastseen when all directories have been checked.

Performance improvements:
- cached_songs_dirty_ is now a multihash where the the path is the key and the songs values, so it does not have to iterate all paths when called from ScanSubdirectory(), only the first time.

- CUE wasn't working, CUE specific fixes:
- FindSongByPath() is changed to FindSongsByPath(), returns all songs with the same path, or the same fingerprint so we don't have to call GetSongsByUrl() again to get all the CUE songs, since this won't work when the path is changed and songs found by fingerprint.
- ParserBase::LoadSong() should not return existing songs in the collection when collection_search is set to false. The songs are already found by the watcher.

jonaski:
More bugfixes related to CUE:
- CollectionBackend::GetSongsByFingerprint() was only returning one song, because I had forgotten to remove break in the loop.
- Calculate the cue_added / cue_deleted different when songs are found by fingerprint, since the old cue path never exists.
- Need to iterate all matched songs so that they are removed from t->deleted_songs

jonaski:
I think it's working pretty good now.
Tested renaming, moving many albums out and back in to the collection, moving them between 2 collection directories, they reappear with ratings without duplicates in the DB.
I've only tested basic CUE support, it could still needs more testing, things like losing CUE sheet, and re-adding CUE sheet when the songs are restored and identified by fingerprints.
But I don't use CUE and only have a few albums for testing.

gavinhoward:
Yeah, testing didn't show any issues for me either.

Congrats!

PR #756, opened 2021-09-07
Fix setQuitOnLastWindowClosed called from instance

EmmanuelMess:
A small fix.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -192,7 +192,7 @@ int main(int argc, char *argv[]) {
     }
     return 0;
   }
-  a.setQuitOnLastWindowClosed(false);
+  QGuiApplication::setQuitOnLastWindowClosed(false);
 
 #if defined(USE_BUNDLE) && (defined(Q_OS_LINUX) || defined(Q_OS_MACOS))
   qLog(Debug) << "Looking for resources in" << QCoreApplication::applicationDirPath() + "/" + USE_BUNDLE_DIR;
@@ -272,7 +272,7 @@ int main(int argc, char *argv[]) {
 #endif
   translations->LoadTranslation("strawberry", ":/translations", language);
   translations->LoadTranslation("strawberry", TRANSLATIONS_DIR, language);
-  translations->LoadTranslation("strawberry", a.applicationDirPath(), language);
+  translations->LoadTranslation("strawberry", QCoreApplication::applicationDirPath(), language);
   translations->LoadTranslation("strawberry", QDir::currentPath(), language);
 
 #ifdef HAVE_QTSPARKLE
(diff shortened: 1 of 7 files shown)


jonaski:
Thanks. This seem correct. I didn't even know this function was static because it's not listed as static in the documentation.
But maybe change this to: `QApplication::setQuitOnLastWindowClosed`, that's what we do for the other QApplication functions.

EmmanuelMess:
> Thanks. This seem correct. I didn't even know this function was static because it's not listed as static in the documentation.
> But maybe change this to: `QApplication::setQuitOnLastWindowClosed`, that's what we do for the other QApplication functions.

Thanks! Please approve the workflow.

EmmanuelMess:
Fixed a few more functions in main.

src/main.cpp
```diff
@@ -192,7 +192,7 @@ int main(int argc, char *argv[]) {
     }
     return 0;
   }
-  a.setQuitOnLastWindowClosed(false);
+  SingleApplication::setQuitOnLastWindowClosed(false);
```
jonaski, line 195: QGuiApplication

src/main.cpp
```diff
@@ -272,7 +272,7 @@ int main(int argc, char *argv[]) {
 #endif
   translations->LoadTranslation("strawberry", ":/translations", language);
   translations->LoadTranslation("strawberry", TRANSLATIONS_DIR, language);
-  translations->LoadTranslation("strawberry", a.applicationDirPath(), language);
+  translations->LoadTranslation("strawberry", SingleApplication::applicationDirPath(), language);
```
jonaski, line 275: QCoreApplication

src/main.cpp
```diff
@@ -313,7 +313,7 @@ int main(int argc, char *argv[]) {
 #endif
   QObject::connect(&a, &SingleApplication::receivedMessage, &w, QOverload<quint32, const QByteArray&>::of(&MainWindow::CommandlineOptionsReceived));
 
-  int ret = a.exec();
+  int ret = SingleApplication::exec();
```
jonaski, line 316: QCoreApplication

jonaski:
The following can be changed `QCoreApplication::applicationDirPath` (consistent with other places in the code):
https://github.com/strawberrymusicplayer/strawberry/blob/master/ext/libstrawberry-common/core/workerpool.h#L236
If you grep qApp, you'll se a few more instances.
`qApp->exit()` and `qApp->quit()` can use static `QCoreApplication::exit()` and `QCoreApplication::quit()`

jonaski (changes requested):

src/main.cpp
EmmanuelMess: This is buggy, it doesn't work if I change to `QGuiApplication`.

EmmanuelMess:
Fixed!

jonaski (approved):

PR #765, opened 2021-09-12
Use optional to fix song fields that were positive or -1

EmmanuelMess:
A few fields in songs used the `-1` value to mean that it was not filled. I replace that with std::optional so that it is more clear that the value can be not filled.

### DIFF
--- a/src/core/song.cpp
+++ b/src/core/song.cpp
@@ -164,7 +164,7 @@ struct Song::Private : public QSharedData {
   explicit Private(Source source = Source_Unknown);
 
   bool valid_;
-  int id_;
+  std::optional<uint> id_;
 
   QString title_;
   QString title_sortable_;
@@ -174,10 +174,10 @@ struct Song::Private : public QSharedData {
   QString artist_sortable_;
   QString albumartist_;
   QString albumartist_sortable_;
-  int track_;
-  int disc_;
-  int year_;
-  int originalyear_;
+  std::optional<uint> track_;
+  std::optional<uint> disc_;
+  std::optional<uint> year_;
+  std::optional<uint> originalyear_;
   QString genre_;
   bool compilation_;		// From the file tag
   QString composer_;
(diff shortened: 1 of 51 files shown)


src/core/song.cpp
```diff
@@ -862,26 +868,26 @@ void Song::ToProtobuf(spb::tagreader::SongMetadata *pb) const {
   pb->set_performer(DataCommaSizeFromQString(d->performer_));
   pb->set_grouping(DataCommaSizeFromQString(d->grouping_));
   pb->set_lyrics(DataCommaSizeFromQString(d->lyrics_));
-  pb->set_track(d->track_);
-  pb->set_disc(d->disc_);
-  pb->set_year(d->year_);
-  pb->set_originalyear(d->originalyear_);
+  if(d->track_) pb->set_track(d->track_.value());
```
jonaski, line 871: Please follow the code-style here and all similar lines, space between if and parentheses.
`if (d->track_) pb->set_track(d->track_.value());`

src/core/song.cpp
```diff
@@ -1635,7 +1632,11 @@ void Song::ToXesam(QVariantMap *map) const {
   AddMetadataAsList("xesam:artist", artist(), map);
   AddMetadata("xesam:album", album(), map);
   AddMetadataAsList("xesam:albumArtist", albumartist(), map);
-  AddMetadata("mpris:length", length_nanosec() / kNsecPerUsec, map);
+  if(length_nanosec()) {
+    AddMetadata("mpris:length", length_nanosec().value() / quint64(kNsecPerUsec), map);
+  } else {
```
jonaski, line 1637:
```
}
else {
```
for consistency with the code-style.

jonaski (changes requested):

EmmanuelMess:
This should also fix the regression with unknown length songs.

jonaski:
This might be a good idea, but require a lot of testing.
With these changes 13 unit tests fail:
```

[  FAILED  ] SingleSong.TestUrls (5 ms)
[----------] 12 tests from SingleSong (67 ms total)

[----------] Global test environment tear-down
[==========] 17 tests from 2 test suites ran. (98 ms total)
[  PASSED  ] 4 tests.
[  FAILED  ] 13 tests, listed below:
[  FAILED  ] CollectionBackendTest.AddInvalidSong
[  FAILED  ] SingleSong.GetSongWithNoAlbum
[  FAILED  ] SingleSong.GetAllArtists
[  FAILED  ] SingleSong.GetAllAlbums
[  FAILED  ] SingleSong.GetAlbumsByArtist
[  FAILED  ] SingleSong.GetAlbumArt
[  FAILED  ] SingleSong.GetSongs
[  FAILED  ] SingleSong.GetSongById
[  FAILED  ] SingleSong.FindSongsInDirectory
[  FAILED  ] SingleSong.UpdateSong
[  FAILED  ] SingleSong.DeleteSongs
[  FAILED  ] SingleSong.MarkSongsUnavailable
[  FAILED  ] SingleSong.TestUrls

13 FAILED TESTS

```

jonaski:
It also crashes:

```
22:34:59.644 DEBUG PlaylistBackend:336              Saving playlist 4
22:34:59.645 WARN  ScopedTransaction:35             Rolling back transaction
22:34:59.645 WARN  unknown                          Qt has caught an exception thrown from an event handler. Throwing 
22:34:59.645 WARN  unknown                          exceptions from an event handler is not supported in Qt. 
22:34:59.645 WARN  unknown                          You must not let any exception whatsoever propagate through Qt code. 
22:34:59.645 WARN  unknown                          If that is not possible, in Qt 5 you must at least reimplement 
22:34:59.645 WARN  unknown                          QCoreApplication::notify() and catch all exceptions there. 
22:34:59.645 WARN  unknown                           
terminate called after throwing an instance of 'std::bad_optional_access'
  what():  bad optional access
Aborted (core dumped)
```

EmmanuelMess:
@jonaski the only places where bad_optional_access can occur is when accessing id, this is a special case, because once it is set it should not be unset again, but I must have missed the correct program flow where in the execution the id is set and left too many `id().value()` in code, I would replace all of these with value_or(-1), this however has the problem of being equivalent to `static_cast<uint>(-1)` in cases where id is empty. What is the maximum id feasible?

jonaski:
I think uint is fine, but we need to check everywhere that id is actual set before using value().

EmmanuelMess:
I will try to fix these in multiple PRs.

PR #770, opened 2021-09-14
Replace static_casts with dynamic_casts

EmmanuelMess:


src/covermanager/coverproviders.cpp
```diff
@@ -134,7 +134,7 @@ void CoverProviders::RemoveProvider(CoverProvider *provider) {
 
 void CoverProviders::ProviderDestroyed() {
 
-  CoverProvider *provider = static_cast<CoverProvider*>(sender());
+  CoverProvider *provider = dynamic_cast<CoverProvider*>(sender());
```
jonaski, line 137: `dynamic_cast` does not work here when `sender()` is destroyed, that's the reason `qobject_cast` wasn't used here in the first place. This will crash.

src/device/devicemanager.cpp
```diff
@@ -221,7 +221,7 @@ void DeviceManager::ListerClosed() {
 
 void DeviceManager::DeviceDestroyed() {
 
-  ConnectedDevice *device = static_cast<ConnectedDevice*>(sender());
+  ConnectedDevice *device = dynamic_cast<ConnectedDevice*>(sender());
```
jonaski, line 224: `dynamic_cast` does not work here when `sender()` is destroyed, that's the reason `qobject_cast` wasn't used here in the first place. This will crash.

src/lyrics/lyricsproviders.cpp
```diff
@@ -133,7 +133,7 @@ void LyricsProviders::RemoveProvider(LyricsProvider *provider) {
 
 void LyricsProviders::ProviderDestroyed() {
 
-  LyricsProvider *provider = static_cast<LyricsProvider*>(sender());
+  LyricsProvider *provider = dynamic_cast<LyricsProvider*>(sender());
```
jonaski, line 136: `dynamic_cast` does not work here when `sender()` is destroyed, that's the reason `qobject_cast` wasn't used here in the first place. This will crash.

jonaski (changes requested):

src/core/mergedproxymodel.cpp
```diff
@@ -219,7 +219,7 @@ void MergedProxyModel::SubModelAboutToBeReset() {
 
 void MergedProxyModel::SubModelResetSlot() {
 
-  QAbstractItemModel *submodel = static_cast<QAbstractItemModel*>(sender());
+  QAbstractItemModel *submodel = dynamic_cast<QAbstractItemModel*>(sender());
```
jonaski, line 222: I think `qobject_cast` can be used here, but we should test.

jonaski:
Using dynamic_cast where sender() has been destroyed does not work, those needs to be reverted.
I'm unsure if it's a good idea to switch to `dynamic_cast` for `QEvents`. Qt itself uses `static_cast`, and I also found this:
https://stackoverflow.com/questions/53413913/should-i-check-qevent-dynamic-cast-result-if-i-already-checked-event-type
Unsure about `QFutureWatcher` too.

jonaski (changes requested):

PR #773, opened 2021-09-19
Store Tidal MPEG-DASH file in data uri

TpmKranz:
Apparently, Clementine and Strawberry can process data URIs just fine, so creating physical files with MPEG-DASH descriptions for Tidal streams can be avoided.

I guess the check for the MPD tag is premature generalization, but that's for y'all to decide.

I'd also like to hear your opinion on whether to use the base64-encoded manifest string, which would get rid of the QByteArray to QString conversion but would force the player to decode the base64 again and use a bit more space than the current percent-encoded path. OTOH, as it stands now, the player would need to do decoding anyway, just not base64 but percent-encoding.

If both of these simplifications were applied, lines 215 to 221 would boil down to a single `url.setPath`.

### DIFF
--- a/src/tidal/tidalstreamurlrequest.cpp
+++ b/src/tidal/tidalstreamurlrequest.cpp
@@ -210,29 +210,10 @@ void TidalStreamURLRequest::StreamURLReceived() {
 
     QXmlStreamReader xml_reader(data_manifest);
     if (xml_reader.readNextStartElement()) {
-
-      QString filepath = QStandardPaths::writableLocation(QStandardPaths::CacheLocation) + "/tidalstreams";
-      QString filename = "tidal-" + QString::number(song_id_) + ".xml";
-      if (!QDir().mkpath(filepath)) {
-        Error(QString("Failed to create directory %1.").arg(filepath), json_obj);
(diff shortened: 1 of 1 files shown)


jonaski:
Thanks. This is probably correct. I have limited knowledge of MPEG-DASH.

TpmKranz:
It appears that Qt6 doesn't like to compare bare strings with `QStringRef`s, so I've simplified it as per my initial remarks.

>I have limited knowledge of MPEG-DASH.
This has not much to do with MPEG-DASH in particular, just the way the player “reads” the input file for a stream. My change only makes it so that the file doesn't need to be written out to a filesystem location.

jonaski:
> It appears that Qt6 doesn't like to compare bare strings with `QStringRef`s, so I've simplified it as per my initial remarks.
> 
Qt 6 doesn't use `QStringRef`, it uses `QStringView`. But both can use `toString()`.
See: https://doc-snapshots.qt.io/qt6-dev/qxmlstreamreader.html#name

jonaski (approved):

jonaski:
Thanks!

PR #797, opened 2021-10-15
Fix filter menu button arrow overlap

buckmelanoma:
Change collection filter menu button's popupMode to MenuButtonPopup to prevent arrow overlapping with button icon


jonaski:
Thanks @buckmelanoma

foss-:
fix verified, thanks so much!

PR #816, opened 2021-10-30
Add icon names for missing Tools menu entries

buckmelanoma:
I noticed the abort_collection_scan, console and sidebar menu entries were missing matching icons.  This change adds appropriate icons for each action.

icns2


jonaski:
It does not work when not using system icons so I think we need to find some icons to include.

buckmelanoma:
Above commit changes icon names to reuse the built-in icons when system icons is disabled.  Looks good in both system and non-system modes:

Built-in icons:
builtin

System icons:
breeze

jonaski:
Thanks.
I sqashed the commits in https://github.com/strawberrymusicplayer/strawberry/commit/20c2225d8f00247fe5ab6495dabbcad603f8f8ab

PR #823, opened 2021-11-07
Prevent playlist filtering when filter toolbar is hidden, clear when toolbar is disabled

buckmelanoma:
This is a followup to #809 

This commit prevents the playlist from being filtered when the playlist filter toolbar is disable in settings.  It also clears any existing filter text as the playlist filter toolbar is set to disabled.

Peek 2021-11-07 10-57


jonaski:
I manually rebased this to fix the formatting.
Merged in https://github.com/strawberrymusicplayer/strawberry/commit/7b8265d4a3ad400c441bdd455e1c4c11e2ad32ef
By the way: I think you're e-mail address is incorrectly set, the commit isn't linked to your account properly.

Issue #821, opened 2021-11-02
Playlist export does not URL encode paths


eduarrrd:
**Describe the bug**

When exporting to xspf or m3u (haven't tried the others) the resulting file paths are not URL encoded.

For xspf this is explicitly called out as a requirement here: https://xspf.org/xspf-v1.html#rfc.section.6.2 (and implicitly at 4.1.1.2.5 by specifying the as URIs).

M3U is a bit murkier though presumably the `U` stands for URL (c.f. VLC devs struggling with this in https://code.videolan.org/videolan/vlc/-/issues/19567).

**To Reproduce**
1. Create a playlist where a track has a bracket (`[`) in its file path.
2. Export to .xspf
3. Visually inspect the file

Alternatively,

3. Open .xspf in VLC
4. See it not playing that track

**Expected behavior**
Paths in playlist exports are URL encoded, and thus conforming to spec.

**Screenshots:**

**System Information:**
 - Operating system: Linux
 - Strawberry Version: 0.9.2

**Additional context**
Note that most players, including VLC, do a lot of the work themselves. For example, spaces (` `) are interpreted correctly despite not being URL encoded. For VLC in particular these helpful assists stop at brackets (`[`).

I would suggest to also take into account that Strawberry is exporting playlists as UTF-8 anyway and only URL encode RFC 3986 section 2.2 Reserved Characters rather than "anything non ascii". I.e. please leave non-latin script readable :-).

buckmelanoma:
Line 157 of xspfparser.cpp is responsible for creating the location tag:

`QString filename_or_url = URLOrFilename(song.url(), dir, path_type).toUtf8();`

buckmelanoma:
This change produces an escaped location tag but strangely does not escape brackets:

`QString filename_or_url = QUrl(URLOrFilename(song.url(), dir, path_type)).toEncoded(QUrl::FullyEncoded);`

buckmelanoma:
PR #826

jonaski:
Looks to me like VLC is using HTML entities, not percent encoding for XSPF playlists. Is that correct?
See: https://code.videolan.org/videolan/vlc/-/blob/master/modules/misc/playlist/xspf.c#L47 and 
https://code.videolan.org/videolan/vlc/-/blob/master/src/text/strings.c#L289
Looking at the specifications, I'm not sure I understand what's correct.
[...]

PR #826, opened 2021-11-10
Use toPercentEncoding for XSPF playlist writer

buckmelanoma:
Fix for #821

This change adds percent encoding to the location tag on exported XSPF playlists.  Tested encoded XSPF with VLC.

toPercentEncoding has built-in UTF-8 conversion per the docs, so no need to do it twice:
https://doc.qt.io/qt-5/qurl.html#toPercentEncoding


buckmelanoma:
Will reopen in a new PR

PR #827, opened 2021-11-10
Use toPercentEncoding for XSPF playlist writer

buckmelanoma:
Fix for #821

This change adds percent encoding to the location tag on exported XSPF playlists. Tested encoded XSPF with VLC.

toPercentEncoding has built-in UTF-8 conversion per the docs, so no need to do it twice:
https://doc.qt.io/qt-5/qurl.html#toPercentEncoding

### DIFF
--- a/src/playlistparsers/xspfparser.cpp
+++ b/src/playlistparsers/xspfparser.cpp
@@ -154,7 +154,7 @@ void XSPFParser::Save(const SongList &songs, QIODevice *device, const QDir &dir,
 
   StreamElement tracklist("trackList", &writer);
   for (const Song &song : songs) {
-    QString filename_or_url = URLOrFilename(song.url(), dir, path_type).toUtf8();
+    QString filename_or_url = QUrl::toPercentEncoding(URLOrFilename(song.url(), dir, path_type));
 
     StreamElement track("track", &writer);
     writer.writeTextElement("location", filename_or_url);


jonaski:
Pretty sure this will break loading back playlists in Strawberry.
In `ParserBase::LoadSong` we use QUrl::fromUserInput() for stream URL's and `QUrl::toLocalFile()` for file. `QUrl::toPercentEncoding()` encodes even `/`.
If location is just a filename we use `QUrl::fromLocalFile()`, I don't think that handles everything encoded by `QUrl::toPercentEncoding()`.

We should probably do it in `ParserBase::URLOrFilename` so that it applies to all playlist formats, perhaps something like this:

```
diff --git a/src/playlistparsers/parserbase.cpp b/src/playlistparsers/parserbase.cpp
index 024fdf6d..983d1567 100644
--- a/src/playlistparsers/parserbase.cpp
+++ b/src/playlistparsers/parserbase.cpp
@@ -103,9 +103,9 @@ Song ParserBase::LoadSong(const QString &filename_or_url, const qint64 beginning
 
 QString ParserBase::URLOrFilename(const QUrl &url, const QDir &dir, Playlist::Path path_type) {
 
-  if (!url.isLocalFile()) return url.toString();
+  if (!url.isLocalFile()) return url.toEncoded(QUrl::FullyEncoded);
 
-  const QString filename = url.toLocalFile();
+  const QString filename = url.toEncoded(QUrl::FullyEncoded|QUrl::PreferLocalFile);
 
   if (path_type != Playlist::Path_Absolute && QDir::isAbsolutePath(filename)) {
     const QString relative = dir.relativeFilePath(filename);

```

buckmelanoma:
Doh!  I knew I missed something.  I'll convert this back to a draft and work on developing this a bit more.

PR #831, opened 2021-11-18
Use new MusicBrainz icon on Edit Track Tag Dialog

buckmelanoma:
Currently the Edit Track Information is using the older MusicBrainz icon for the Tag feature.  This PR replaces the icon with the recently added MB icon.

Screenshot_20211117_161529


jonaski:
I prefer to keep the existing image for the button

PR #837, opened 2021-11-25
Update strawberry.css

dylanflogaus:
Added white color to QMenu since I am trying to use a black background.

### DIFF
--- a/data/style/strawberry.css
+++ b/data/style/strawberry.css
@@ -64,7 +64,7 @@ macos {
 
 macos QMenu {
   font-size: 13pt;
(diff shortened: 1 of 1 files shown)


jonaski:
I don't think we should be hard-coding colors in the CSS unless it's strictly necessary.
What are you trying to change here? The text color of the context menus?

dylanflogaus:
Grey would let any background show the text :D please help, I like to keep the background nice and dark...

dylanflogaus:
Looks great now, thank you very much!

Issue #410, opened 2020-04-20
Freeze when attempting to access systemd managed disconnected external device


DAMO238:
**Describe the bug**
When strawberry tries to change song, it stops drawing the window and the system tray icon cannot be interacted with. When running in terminal, you get this output during the crash:

```
16:18:14.745 DEBUG PlaylistBackend:332              Saving playlist 18
16:18:16.601 ERROR logging:71(GLib)                 Source ID 24 was not found when attempting to remove it
Error creating proxy: Error calling StartServiceByName for org.gtk.vfs.UDisks2VolumeMonitor: Timeout was reached (g-io-error-quark, 24)
16:18:26.011 INFO  DeviceManager:455                Device added: "Gio//btrfs/119535599616" ("Gio//btrfs/119535599616")
16:18:26.012 INFO  DeviceManager:455                Device added: "Gio//msdos/305569792" ("Gio//btrfs/119535599616", "Gio//msdos/305569792")
16:18:26.013 INFO  DeviceManager:455                Device added: "Gio//btrfs/1000203816960" ("Gio//btrfs/1000203816960", "Gio//btrfs/119535599616", "Gio//msdos/305569792")
16:19:13.600 DEBUG Database:487                     Starting database integrity check
16:19:13.611 DEBUG Database:548                     Starting database backup
16:20:01.286 WARN  GioLister::DeviceInfo:453        Error getting filesystem info for /home/damien/Games: No such device
16:20:01.286 INFO  DeviceManager:455                Device added: "Gio///0" ("Gio///0", "Gio//btrfs/1000203816960", "Gio//btrfs/119535599616", "Gio//msdos/305569792")

```

**To Reproduce**
Use an external drive with the following fstab (specifically the last entry, but I pasted the entire thing in case it is something else):
```
# Static information about the filesystems.
# See fstab(5) for details.

# <file system> <dir> <type> <options> <dump> <pass>
# /dev/sdb1 LABEL=root
UUID=2a72b552-9914-4bc4-90b7-7cae42309be7	/         	btrfs     	rw,nodev,relatime,ssd,discard,space_cache,subvolid=256,subvol=/__current/root,subvol=__current/root	0 0
UUID=2a72b552-9914-4bc4-90b7-7cae42309be7	/run/btrfs-root         	btrfs     	rw,nodev,relatime,ssd,discard,space_cache,nosuid,noexec	0 0

# /dev/sdb2
[...]

PR #870, opened 2022-01-09
GIOLister: ignore system mounts as defined by GIO

WGH-:
Strawberry has some heuristics to exclude things like the root mount, `/boot`, `tmpfs`, etc. from the devices list.

However, it's somewhat incomplete. GIO (GLib component) has more complete definition of "internal system mounts" that should not be presented to the user.

Additionally, don't try to query filesystems free/total size if it's internal, as that triggers automounting for autofs filesystems (#410).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,6 +117,9 @@ endif()
 pkg_check_modules(GLIB REQUIRED glib-2.0)
 pkg_check_modules(GOBJECT REQUIRED gobject-2.0)
 pkg_check_modules(GIO REQUIRED gio-2.0)
+if(UNIX)
+  pkg_check_modules(GIO_UNIX gio-unix-2.0)
+endif()
 pkg_check_modules(LIBCDIO libcdio)
 pkg_check_modules(GSTREAMER gstreamer-1.0)
 pkg_check_modules(GSTREAMER_BASE gstreamer-base-1.0)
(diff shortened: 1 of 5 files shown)


WGH-:
This thing is still messy, though. I couldn't figure out how to exclude unmounted `/boot` from the device list (and I spent quite a bit time on it). Somehow none of the `GVolume` properties had a full `/boot` path that I could pass into `g_unix_*` functions.

CMakeLists.txt
```diff
@@ -117,6 +117,9 @@ endif()
 pkg_check_modules(GLIB REQUIRED glib-2.0)
 pkg_check_modules(GOBJECT REQUIRED gobject-2.0)
 pkg_check_modules(GIO REQUIRED gio-2.0)
+if(UNIX)
+    pkg_check_modules(GIO_UNIX gio-unix-2.0)
```
jonaski, line 121: indent 2 spaces

src/device/giolister.cpp
```diff
@@ -25,6 +25,9 @@
 #include <glib.h>
 #include <glib-object.h>
 #include <gio/gio.h>
+#ifdef HAVE_GIO_UNIX
+#include <gio/gunixmounts.h>
```
jonaski, line 29: `#  include <gio/gunixmounts.h>`

src/device/giolister.cpp
```diff
@@ -434,6 +439,8 @@ void GioLister::DeviceInfo::ReadMountInfo(GMount *mount) {
   mount_ptr.reset_without_add(mount);
   if (!mount) return;
 
+
```
jonaski, line 442: Remove empty lines

jonaski:
Thanks. Looks good to be except for some formatting.

jonaski (changes requested):

src/device/giolister.cpp
WGH-: I assumed you meant two whitespaces here as well

jonaski (approved):

src/device/giolister.cpp
jonaski: Yes

PR #879, opened 2022-01-22
subsonic configuration: use the entered password when clicking Test

cquike:
This fixes a bug in which the password used when clicking in Test was the previously saved password, not the one just entered now.


jonaski:
Thanks!

PR #906, opened 2022-02-24
request 569 - left click selects metadata to modify with F2.

jakubbalas:
https://forum.strawberrymusicplayer.org/topic/569/left-click-selecting-the-column


jakubbalas:
Here are more details of what is the desired outcome
https://forum.strawberrymusicplayer.org/topic/569/left-click-selecting-the-column

Caveats:
I'm not a Cpp dev so had hard time to test it :/  

I didn't test smart playlists

PR #913, opened 2022-03-11
Devcontainer image for vscode

renn0xtek9:
This uses the VSCode feature of development container. 
I think this is usefull also for people who want to develop natively on Ubuntu / KDE Neon, so that all the dependencies are listed and they don't have to google for it


jonaski:
Thanks but I don't think we really need this. I wouldn't mind adding this if more contributors were using it, but right now I think it's just you.

PR #915, opened 2022-03-13
Add "save all playlist" feature

renn0xtek9:
This adds a feature that will save all playlists to m3u in `/home/foo/Playlists` with relative path to the songs. 
This is very usefull if you want to copy paste all you music to e.g. you phone without loosing your playlists.

PR #916, opened 2022-03-13
Add a formatting script and apply the formatting across the repo

renn0xtek9:
This adds a `format-all.sh` that format all source code across the repo.
I suggest to use this as single source of truth for code formatting. This will prevent issues with various settings across different user IDE's.

The PR is obviously huge since the whole repo was formatted accordingly. But there is no code modification whatsoever.
@jonaski thanks for your quick review of https://github.com/strawberrymusicplayer/strawberry/pull/915 . I suggest to merge this present PR before so that all the formatting accross the repo is aligned with the `.clang-format` file


jonaski:
I have never fully used the clang format configuration, I only use it for some guidance, not to re-format all the sources as there are some exceptions to the configuration.
The following differences from the clang-format configuration are intentional, I didn't figure out how to configure clang format to leave them like that, these are:

- Spaces between code and braces in functions, I prefer to keep it like that for readability.
- Although I prefer pointer and reference alignment right, I don't think this should apply when there is no variable name involved in functions, or in casts. 
- There are some places where I've used one-line if statements, like for simply things return early in a function. I prefer to use braces if the statements are split.

3rdparty/getopt should not be reformatted.

jonaski:
If you figure out the clang format settings, it's appreciated. Open a new PR with clang-format changes.
For reformatting all sources I usually use:
`find . -iname '*.h' -o -iname '*.cpp' | xargs clang-format -i`
Simpler than your script.

renn0xtek9:
I could not find an option to do that within clang-format so I add a `sed` command in the format all script to add the new line between closing curly braces and end of code block. It is not really nice, but I don't have a better idea. 
With this addition and the idea that we could use it in the future to format also non C++ file, I would rather keep this script for now?
Also what do you think about making those changes in the `.clang-format` ? (Not included in this PR)
```
SortIncludes: true
SortUsingDeclarations: true
```
I find it nicer when sorted properly.

jonaski:
It's still formatting too much. Maybe we should just remove the .clang-format file if it's too confusing.

renn0xtek9:
I mean once this is formatted, it should be done once for all? That would make subsequent work much easier. (no more comment on formatting in PR) or ?

Issue #926, opened 2022-03-25
Corrupted backgroung image strawbs.png if dark theme is enabled in KDE


adem4ik:
**Describe the bug**
Background image strawbs.png gets corrupted if KDE has dark theme enabled

**To Reproduce**
- Switch to dark theme in KDE
- Set playlist's background image to `A taste of Strawbs`

**Expected behavior**
Same quality of background image as in light theme

**Screenshots:**
![image](https://user-images.githubusercontent.com/4707112/160150378-fd52f421-958a-4935-a9b4-e4e8c9646f71.png)

**System Information:**
 - Operating system: Manjaro with KDE Plasma 5.24.3
 - Strawberry Version: 1.0.3-1-g49726701 (latest git version from AUR)

**Additional context**
I guess the problem comes from the transparency of the original PNG - https://github.com/strawberrymusicplayer/strawberry/blob/master/data/pictures/strawbs.png After deleting transparency from this PNG, I got the following result: 
![image](https://user-images.githubusercontent.com/4707112/160153208-278a1af0-8cbb-40f9-9711-67f42c10c235.png)

And btw it looks corrupted on GitHub too, if it has dark theme enabled:
![image](https://user-images.githubusercontent.com/4707112/160153943-2447f6b9-ba09-4554-b14b-4692ef672662.png)

jonaski:
I don't think there is an fix for this, the image doesn't work with a black background.

PR #929, opened 2022-03-27
Rework strawbs.png to make it suitable for dark theme

adem4ik:
Fixes https://github.com/strawberrymusicplayer/strawberry/issues/926

Things done:
- Corrected color balance from mostly red to partially red / partially grey
- Upscaled using waifu2x including noise reduction and then downscaled back to original size to recover picture
- Optimized filesize via TinyPNG & FileOptimizer

2022-03-28_03-22_light_comp
2022-03-28_03-24_dark_comp


jonaski:
I'll test this out but I'm not sure it's visible enough to be usable with a dark theme.

adem4ik:
I have another version. It was dropped cause it looks too dark in light theme. But works better in dark theme: [strawbs_v2.zip](https://github.com/strawberrymusicplayer/strawberry/files/8422037/strawbs_v2.zip)

Here is comparison (v1, left - from my PR, v2 is from this message): 
image
image

Let me know, if I should remake the PR.

PR #930, opened 2022-03-29
Rework Strawberry icon

adem4ik:
Changes:
- Bigger size of Strawberry icon to match with other system icons
- Corrected rotation (see below):

2022-03-29_19-12_problem
2022-03-29_19-18_move

A few comparisons:
- Taskbar
2022-03-30_00-31_comp1_taskpanel
- Tray
2022-03-30_00-31_comp_tray

My source image which is upscaled and fixed version of original /data/icons/full/strawberry.png: [strawberry_1856.zip](https://github.com/strawberrymusicplayer/strawberry/files/8374933/strawberry_1856.zip)


tybug:
personally I prefer the "incorrect" rotation, this version feels unnaturally flat. Not my place to judge though.

jonaski:
Thanks, but I prefer keeping the icon to the original version.

PR #982, opened 2022-06-15
README: Exclude unsupported repos in the repology badge

WOnder93:
Showing them provides little value, as they are no longer maintained
and thus expected to contain outdated packages.


jonaski:
Good idea, I didn't know about this option. Thanks.

Issue #274, opened 2019-09-30
Scrobbling Album Artist instead of Artist


beescopy:
**Is your feature request related to a problem? Please describe.**
Songs are normally tagged like this (this is for example used by the musicbrainz database):
Title = "SongTitle"
Album = "AlbumName"
Artist = "Artist1 feat. Artist2 and Artist3"
Album Artist = "Artist1"

**Last.FM gets confused when "Artist1 feat. Artist2 and Artist3" is scrobbled as Artist.**
Most times the song can't be linked to an artist page, except when all artists in this exact configuration are known to have been working together on full albums or many songs, or when the scrobbled song is indeed a single.

**The result?**
Missing Coverart and weird listening statistics.
Listening to only one Album might result in "This week you mostly listened to:
'Artist1'
'Artist1 feat. Artist2 and Artist3'
'Artist1 feat. Artist4'
'Artist1 and Artist5'
'Artist1, Artist6 & Artist7'"
instead of just:
'Artist1'

**Describe the solution you'd like**
An option to send the Album Artist information as Artist would resolve this issue.

**Additional context**
A simple checkbox "Prefer Album Artist over Artist" in the Scrobbler-Settings would be very helpful.

**thank you.**

jonaski:
It's almost as if last.fm ignores albumArtist completely.

Issue #273, opened 2019-09-29
Send track artist to listenbrainz


guedressel:
**Describe the bug**
In listenbrainz I find the "Album artist" information tracked, tracking the "Artist" information might improve accuracy for listenbrainz. On libre.fm "Artist" gets already tracked instead of "Album Artist".  
Example: Album artist: "Various Artists", (Track) Artist: "Susanne Blech"

**To Reproduce**
Activate listenbrainz tracking. Play a track with different information in "Album Artist" and "Artist" meta tags. Check listenbrainz logs.

**Expected behavior**
Having "Susanne Blech" tracked, instead of "Various Artists".

**System Information:**
 - Arch
 - Strawberry 0.6.3

**Additional context**
I'm enjoying strawberry a lot! Thank you.

jonaski:
Last.fm and libre.fm has it's own parameter for album artist that is sent if album artist is set for the song.
But listenbrainz only have "artist_name", I think the most accurate to send in that case is album artist, since that's what identifies what release it is that is sent in the other parameter "release_name".
But I made an exception, if album artist is set to "various artists" case-insensitive it will use the artist instead.

guedressel:
Thanks! That was a quick fix.

Just began to look into the listenbrainz API myself:
https://listenbrainz.readthedocs.io/en/production/dev/json.html#json-doc

Do you check for *_mbid musicbrainz metatags?
In my collection I have a lot of musicbrainz tags - therefore listenbrainz would easily pick up the recording / release information if strawberry would submit those too.

What do you think about integrating musicbrainz specific tags to the listenbrainz submissions?

jonaski:
It can be done, but it involves a lot more work. We have to create fingerprints for each song then. The best way to do it I think is to calculate the fingerprint at the same time as scanning the collection, then store it in the DB. Then we can simply request the mbid from acoustid.org (we already do that when requesting tags from musicbrainz) before scrobbling a song.

guedressel:
What if a track *already contains* such musicbrainz tags - do you really need to (re-)calculate the musicbrainz fingerprints?

jonaski:
No, but I'm not sure how to read mbid from tags, is it a known tag both in Vorbis comments and ID3v2, does all audio formats support it?

jonaski:
Looks like there is some info here https://picard.musicbrainz.org/docs/mappings/

guedressel:
These are the tags I found in [Picard](https://picard.musicbrainz.org/).  
[...]

PR #989, opened 2022-06-29
Add support for `prefer album artist` option to ListenBrainz

staticssleever668:
Reference #274.

CC #273.

### DIFF
--- a/src/scrobbler/listenbrainzscrobbler.cpp
+++ b/src/scrobbler/listenbrainzscrobbler.cpp
@@ -46,6 +46,7 @@
 #include "core/timeconstants.h"
 #include "core/logging.h"
 #include "internet/localredirectserver.h"
+#include "settings/scrobblersettingspage.h"
 
 #include "audioscrobbler.h"
 #include "scrobblerservice.h"
@@ -76,7 +77,8 @@ ListenBrainzScrobbler::ListenBrainzScrobbler(Application *app, QObject *parent)
       submitted_(false),
       scrobbled_(false),
       timestamp_(0),
-      submit_error_(false) {
+      submit_error_(false),
+      prefer_albumartist_(false) {
 
   refresh_login_timer_.setSingleShot(true);
   QObject::connect(&refresh_login_timer_, &QTimer::timeout, this, &ListenBrainzScrobbler::RequestNewAccessToken);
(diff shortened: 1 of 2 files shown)


src/scrobbler/listenbrainzscrobbler.cpp
```diff
@@ -429,7 +433,7 @@ void ListenBrainzScrobbler::UpdateNowPlaying(const Song &song) {
   title = title.remove(Song::kTitleRemoveMisc);
 
   QJsonObject object_track_metadata;
-  if (song.albumartist().isEmpty() || song.albumartist().compare(Song::kVariousArtists, Qt::CaseInsensitive) == 0) {
```
jonaski, line 432: albumartist can be empty, so artist needs to be used for artist_name in that case.

src/scrobbler/listenbrainzscrobbler.cpp
```diff
@@ -552,7 +556,7 @@ void ListenBrainzScrobbler::Submit() {
     QJsonObject object_listen;
     object_listen.insert("listened_at", QJsonValue::fromVariant(item->timestamp_));
     QJsonObject object_track_metadata;
-    if (item->albumartist_.isEmpty() || item->albumartist_.compare(Song::kVariousArtists, Qt::CaseInsensitive) == 0) {
+    if (!prefer_albumartist_ || item->albumartist_.compare(Song::kVariousArtists, Qt::CaseInsensitive) == 0) {
```
jonaski, line 559: Same here

jonaski (changes requested):

src/scrobbler/listenbrainzscrobbler.cpp
staticssleever668: Right, I'm not even sure why I removed this check...

src/scrobbler/listenbrainzscrobbler.h
```diff
@@ -120,6 +120,8 @@ class ListenBrainzScrobbler : public ScrobblerService {
   QTimer timer_submit_;
   bool submit_error_;
 
+  bool prefer_albumartist_;
```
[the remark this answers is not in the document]
staticssleever668: Done.

jonaski (approved):

PR #995, opened 2022-07-15
Send more information to ListenBrainz

staticssleever668:
- Track duration. This is visible in the listen list.
- Track number. This is not used right now as far as I see, but no reason not to send it.
- Player name and version. This might get useful for ListenBrainz developers later. In a way this might be called privacy-invasive (also it should be noted that the digest of the last commit is included in the `version` field).

### DIFF
--- a/src/scrobbler/listenbrainzscrobbler.cpp
+++ b/src/scrobbler/listenbrainzscrobbler.cpp
@@ -21,6 +21,7 @@
 
 #include <algorithm>
 
+#include <QCoreApplication>
 #include <QtGlobal>
 #include <QDesktopServices>
 #include <QVariant>
@@ -447,6 +448,19 @@ void ListenBrainzScrobbler::UpdateNowPlaying(const Song &song) {
 
   object_track_metadata.insert("track_name", QJsonValue::fromVariant(title));
 
+  QJsonObject object_additional_info;
+
+  object_additional_info.insert("duration_ms", song.length_nanosec() / kNsecPerMsec);
+
+  if (const int track = song.track(); track > 0) {
+    object_additional_info.insert("tracknumber", track);
+  }
+
+  object_additional_info.insert("submission_client", QCoreApplication::applicationName());
+  object_additional_info.insert("submission_client_version", QCoreApplication::applicationVersion());
+
+  object_track_metadata.insert("additional_info", object_additional_info);
+
   QJsonObject object_listen;
   object_listen.insert("track_metadata", object_track_metadata);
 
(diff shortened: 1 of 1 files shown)


src/scrobbler/listenbrainzscrobbler.cpp
```diff
@@ -569,6 +583,19 @@ void ListenBrainzScrobbler::Submit() {
...
+    object_additional_info.insert("duration_ms", item->duration_ / kNsecPerMsec);
+
+    if (item->track_ > 0) {
+      object_additional_info.insert("tracknumber", item->track_);
+    }
+
+    object_additional_info.insert("submission_client", QCoreApplication::applicationName());
+    object_additional_info.insert("submission_client_version", QCoreApplication::applicationVersion());
+
+    object_track_metadata.insert("additional_info", object_additional_info);
```
staticssleever668, line 597: In my own code I often put such code into immediately invoked function expressions, usually to make the object `const`. This is not used in C++ often, but maybe you will like it.
```suggestion
    QJsonObject object_additional_info = [&]() {
      QJsonObject object_additional_info;

      object_additional_info.insert("duration_ms", item->duration_ / kNsecPerMsec);

      if (item->track_ > 0) {
        object_additional_info.insert("tracknumber", item->track_);
      }

      object_additional_info.insert("submission_client", QCoreApplication::applicationName());
      object_additional_info.insert("submission_client_version", QCoreApplication::applicationVersion());

      return object_additional_info;
    }();
```

src/scrobbler/listenbrainzscrobbler.cpp
```diff
@@ -447,6 +448,19 @@ void ListenBrainzScrobbler::UpdateNowPlaying(const Song &song) {
...
+  QJsonObject object_additional_info;
+
+  object_additional_info.insert("duration_ms", song.length_nanosec() / kNsecPerMsec);
+
+  if (const int track = song.track(); track > 0) {
+    object_additional_info.insert("tracknumber", track);
+  }
+
+  object_additional_info.insert("submission_client", QCoreApplication::applicationName());
+  object_additional_info.insert("submission_client_version", QCoreApplication::applicationVersion());
```
staticssleever668, line 460: ListenBrainz also has `media_player` and `media_player_version` fields, since they are the same thing I didn't include them. I'm not sure which one of them has higher priority, I'll ask about that on their IRC now.

staticssleever668: I asked on IRC, both should be sent. Example: https://listenbrainz.readthedocs.io/en/latest/users/json.html#brainzplayer-on-the-listenbrainz-website-playing-a-video-from-youtube.

jonaski (approved):

PR #1024, opened 2022-08-29
show icons in fileview

Iridias:
**Whats new:**
Show icons in FileView (again?)

**Problem description:**
Don't know about other systems, but on mine, there are no icons in the fileview.
OS: **Arch Linux**
QT: **6.3.1**
Strawberry: **1.0.8**
strawberry_fileview_broken

**Wild guesses:**
Don't know if that has been the case in the past.
Maybe the implementation of `QFileSystemModel` changed recently to not have an `IconProvider` by default.

**The fix:**
Just setting a `QFileIconProvider` makes the icons work again:
strawberry_fileview_fixed

### DIFF
--- a/src/widgets/fileview.cpp
+++ b/src/widgets/fileview.cpp
@@ -268,9 +268,12 @@ void FileView::showEvent(QShowEvent *e) {
   if (model_) return;
 
   model_ = new QFileSystemModel(this);
+  if(!model_->iconProvider() || model_->iconProvider()->icon(QAbstractFileIconProvider::Folder).isNull()) {
+    qLog(Info) << "detected unsuitable iconProvider - try to fix it...";
+    model_->setIconProvider(new QFileIconProvider());
+  }
 
   model_->setNameFilters(filter_list_);
-  model_->setIconProvider(new QFileIconProvider());
   // if an item fails the filter, hide it
   model_->setNameFilterDisables(false);
 
--- a/src/widgets/fileview.cpp
+++ b/src/widgets/fileview.cpp
@@ -21,6 +21,7 @@
 #include <QWidget>
 #include <QUndoStack>
 #include <QDir>
+#include <QFileIconProvider>
 #include <QFileInfo>
 #include <QFileSystemModel>
 #include <QString>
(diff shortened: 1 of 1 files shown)


src/widgets/fileview.cpp
```diff
@@ -268,6 +270,7 @@ void FileView::showEvent(QShowEvent *e) {
   model_ = new QFileSystemModel(this);
 
   model_->setNameFilters(filter_list_);
+  model_->setIconProvider(new QFileIconProvider());
```
jonaski, line 273: This creates a memory leak. Why is it necessary to construct a new `QFileIconProvider` in the first place? I would think it should be possible to set options on the existing one accessed with `model->iconProvider()`

jonaski:
Icons shows fine in the file view for me with Qt 6.3.1 without this patch. I think this is more likely related to some Qt style settings that you override when you construct a new `QFileIconProvider`

src/widgets/fileview.cpp
Iridias: > This creates a memory leak.

Hm, how is that? :thinking: 
This should only be created once, due to the  `if (model_) return;` a few lines before.
(#noirony: I usually don't do stuff in C++ very often, let alone QT - so maybe I missed something)


> Why is it necessary to construct a new QFileIconProvider in the first place?

A reasonable question. The only Option-Flag I found in the [documentation](https://doc.qt.io/qt-6/qabstractfileiconprovider.html#Option-enum) was `DontUseCustomDirectoryIcons` - and it had no effect.

But thinking about it: it might be a good idea, to check if the current IconProvider is usable, before setting a new one.
Something like `provider->icon(QAbstractFileIconProvider::Folder).isNull()`

...will prepare an update for the PR.
But any other suggestions wouldn't go amiss.

jonaski: > > This creates a memory leak.
> 
> Hm, how is that? thinking This should only be created once, due to the `if (model_) return;` a few lines before. (#noirony: I usually don't do stuff in C++ very often, let alone QT - so maybe I missed something)
> 
Because the memory should be fee'd on exit.
Use `std::unique_ptr`

src/widgets/fileview.cpp
```diff
@@ -44,6 +45,7 @@
 #include "organize/organizeerrordialog.h"
 #include "settings/appearancesettingspage.h"
 
+
```
jonaski, line 48: Please remove

jonaski:
Merged in 7dff6f26bc5cbc9470b2c5553e8fa72333ea9649

PR #1036, opened 2022-09-10
Update de.po

geroyche:
The German translation mostly uses fomal you "Sie", but some spots are inconsistent and use informal "du". I also noticed some typo while I was addressing this. And Token is used in German as well. "Zeichen" is definitely a wrong translation.


jonaski:
Translations needs to be updated through Zanata: https://wiki.strawberrymusicplayer.org/wiki/How_to_translate

geroyche:
Ok

Jonas Kvinge ***@***.***> schrieb am Sa., 10. Sept. 2022,
15:12:

PR #1041, opened 2022-09-16
Update README.md

tvgold42:
now mentions the requirement to be on Linux to compile the application, should just be a short time fix until the CMake deps can be updated/adjusted to not require pkg-config, a Linux only dependacy


jonaski:
pkg-config is not a Linux only dependency, it can be used on all Unix based systems and Windows too. Strawberry uses CMake and pkg-config to build with Visual Studio, see https://github.com/strawberrymusicplayer/strawberry-msvc-dependencies

PR #1082, opened 2022-12-04
Allow album artists to be "Various Artists"

RaylaLe100:
Very simple change and I understand that it might not be consensual (hence the PR, we can discuss this here). I scrobble album artist for albums marked under "various artists", and a lot of people do judging by lastfm history (example: https://www.last.fm/music/Various+Artists ).

Before I start arguing aimlessly, is there any specific reason for this limitation? Any usecase where people wouldn't want to have their various artist albums merged on lastfm? If there isn't we can merge the PR as-is, but otherwise I think adding this as a choice for the user would be a good idea (as like a sub-checkbox under `prefer_albumartist_`.

### DIFF
--- a/src/scrobbler/scrobblingapi20.cpp
+++ b/src/scrobbler/scrobblingapi20.cpp
@@ -467,7 +467,7 @@ void ScrobblingAPI20::UpdateNowPlaying(const Song &song) {
     params << Param("album", album);
   }
 
-  if (!prefer_albumartist_ && !song.albumartist().isEmpty() && song.albumartist().compare(Song::kVariousArtists, Qt::CaseInsensitive) != 0) {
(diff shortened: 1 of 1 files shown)


jonaski:
This is only updating it in one place for now playing, we should do the same for "artist". Also, `Submit()`, `SendSingleScrobble()` and `Love()`. Perhaps also do the same for listenbrainz.
I suggest to grep through the entire sources to the catch all the places where `kVariousArtists` is used.

jonaski:
You can also remove `Song::kVariousArtists`, as it it is only used for the scrobbler.

thekingrat:
I was just looking at this work for the last.fm scrobbler, so if this goes in I won't have to create & maintain a fork!

PR #1159, opened 2023-03-11
Add icons to the collection menu button

buckmelanoma:
Adds icons to the menu entries that don't current have one assigned.  Ends up looking like this:

image


jonaski:
Most of these icons don't exist in qt resources (https://github.com/strawberrymusicplayer/strawberry/blob/master/data/icons.qrc).

buckmelanoma:
My bad.  I'll resubmit when I have this working on all the platforms.

PR #1161, opened 2023-03-23
Fix album art rendering on High DPI displays

FletcherD:
Rendering of album art does not take advantage of High DPI displays currently, leading to blurry album art on high density displays in every location.

This change creates album art QPixmaps with the appropriate pixel density and devicePixelRatio of the current widget fixing this issue.

### DIFF
--- a/src/context/contextalbum.cpp
+++ b/src/context/contextalbum.cpp
@@ -63,7 +63,7 @@ ContextAlbum::ContextAlbum(QWidget *parent)
   cover_loader_options_.desired_height_ = width();
   cover_loader_options_.pad_output_image_ = true;
   cover_loader_options_.scale_output_image_ = true;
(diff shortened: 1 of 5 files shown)


FletcherD:
Apologies for this tangent, but I couldn't quite find a better place to ask this.

I'd also like to create another pull request to allow the user to prioritize album art found in the local image file, rather than art embedded in the audio file. Do you have any advice on where I need to make changes?

jonaski:
> Apologies for this tangent, but I couldn't quite find a better place to ask this.
> 
> I'd also like to create another pull request to allow the user to prioritize album art found in the local image file, rather than art embedded in the audio file. Do you have any advice on where I need to make changes?

TryLoadImage in https://github.com/strawberrymusicplayer/strawberry/blob/master/src/covermanager/albumcoverloader.cpp

We could pass it through as part of AlbumCoverLoaderOptions

jonaski:
Thanks, merged https://github.com/strawberrymusicplayer/strawberry/commit/bb43cc63ec71a203623caaad34e49f4d75b46023

PR #1168, opened 2023-03-28
QobuzRequest: Handle multiple discs in integration

gaganpreet:
Albums with multiple discs are not being handled which results in tracks being listed out of order. If the disc (media_number) is present in JSON, we use it to set the disc in the Song object.

### Before
Before

### After
After

### DIFF
--- a/src/qobuz/qobuzrequest.cpp
+++ b/src/qobuz/qobuzrequest.cpp
@@ -1079,11 +1079,16 @@ void QobuzRequest::ParseSong(Song &song, const QJsonObject &json_obj, const Arti
 
   QString title = json_obj["title"].toString();
   int track = json_obj["track_number"].toInt();
+  int disc = 1;
   QString copyright = json_obj["copyright"].toString();
   qint64 duration = json_obj["duration"].toInt() * kNsecPerSec;
   //bool streamable = json_obj["streamable"].toBool();
   QString composer;
   QString performer;
+
+  if (json_obj.contains("media_number")) {
+      disc = json_obj["media_number"].toInt();
+  }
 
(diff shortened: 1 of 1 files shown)


src/qobuz/qobuzrequest.cpp
```diff
@@ -1079,12 +1079,17 @@ void QobuzRequest::ParseSong(Song &song, const QJsonObject &json_obj, const Arti
 
   QString title = json_obj["title"].toString();
   int track = json_obj["track_number"].toInt();
+  int disc = 1;
```
jonaski, line 1082: This should be `0` if Qobuz does not return disc, so the collection model does not show disc in the "Album - Disc" grouping, unless it's an album with multiple discs.

src/qobuz/qobuzrequest.cpp
```diff
@@ -1079,12 +1079,17 @@ void QobuzRequest::ParseSong(Song &song, const QJsonObject &json_obj, const Arti
...
   QString copyright = json_obj["copyright"].toString();
   qint64 duration = json_obj["duration"].toInt() * kNsecPerSec;
   //bool streamable = json_obj["streamable"].toBool();
   QString composer;
   QString performer;
 
+  if (json_obj.contains("media_number")) {
+      disc = json_obj["media_number"].toInt();
```
jonaski, line 1090: Use 2 space indent, not 4.

jonaski (changes requested):

jonaski:
Merged https://github.com/strawberrymusicplayer/strawberry/commit/c1d0702b4d68c30eca2473606b17cbdc2ccfe600

PR #1194, opened 2023-05-04
CueParser: Process composer metadata in cue files

RobertAJMarshall:
Use composer as an alternative to songwriter

..and move a misplaced comment


src/playlistparsers/cueparser.cpp
```diff
@@ -48,6 +48,8 @@ const char *CueParser::kIndexRegExp = "(\\d{1,3}):(\\d{2}):(\\d{2})";
 const char *CueParser::kPerformer = "performer";
 const char *CueParser::kTitle = "title";
 const char *CueParser::kSongWriter = "songwriter";
+// compooser may be in cue file and is synonym for songwriter
```
jonaski, line 51: composer

jonaski (changes requested):

src/playlistparsers/cueparser.cpp
RobertAJMarshall: #headdesk ;)

PR #1212, opened 2023-06-06
Add ability to filter playlists by rating

Dakes:
This adds the ability to filter your playlists by rating in two different ways. 
For example you can filter by 0-5 Stars like this:
`rating:5` `rating:>=3` `rating:!=0`
Or use float values from 0-1 with a prefixed "f":
`rating:f1` `rating:>f0.5`

The implementation got a little messy, unfortunately. But it works.


jonaski (changes requested):
Thanks, could you squash the 2 correction commits into the first one, and if you rebase with master it should also fix the CI issue.

Dakes:
I think it is looking good now. ✌️

jonaski (approved):

Dakes:
Thank you. I appreciate your work. Have a nice day.

PR #1237, opened 2023-07-18
Allow forcing GStreamer pipeline to the floating-point data format

LebedevRI:
While audio is most commonly stored in 16/24-bit integer format,
modern audio processing, and modern audio servers
are using floating point integrally, whereas PipeWire internally
essentially always uses 32-bit floating point, as far as i understand.

While it does not really matter that much to us
as long as we simply pass-through the audio (from the decoder to the playback),
the moment we start to mess with it, it does begin to affect us.

As a "contrived" "obscure" example, pick some audio that has large loudness range,
normalize it to `-0 LUFS`, and load a compressor in easyeffects.
The audio will be distorted, there will be clipping.
Looking at the pipeline dumps, ReplayGain was already handling this correctly:
[pipeline-replaygain.pdf](https://github.com/strawberrymusicplayer/strawberry/files/12088865/pipeline-replaygain.pdf)

Here i propose two things:
* Fix: Always perform EBU R 128 loudness normalization in floating-point
* Feature: subj. Support forcing GStreamer pipeline to always be in floating-point. 

[pipeline-old.pdf](https://github.com/strawberrymusicplayer/strawberry/files/12088778/pipeline-old.pdf)
[pipeline-new-ebur128.pdf](https://github.com/strawberrymusicplayer/strawberry/files/12088922/pipeline-ebur128.pdf)
[pipeline-new-force.pdf](https://github.com/strawberrymusicplayer/strawberry/files/12088923/pipeline-force.pdf)

Thoughts?

### DIFF
--- a/src/engine/enginebase.cpp
+++ b/src/engine/enginebase.cpp
@@ -39,6 +39,7 @@
 
 EngineBase::EngineBase(QObject *parent)
     : QObject(parent),
+      force_floating_point_(false),
       volume_control_(true),
       volume_(100),
       beginning_nanosec_(0),
@@ -166,6 +167,8 @@ void EngineBase::ReloadSettings() {
 
   output_ = s.value("output").toString();
   device_ = s.value("device");
+
+  force_floating_point_ = s.value("force_floating_point", false).toBool();
 
   volume_control_ = s.value("volume_control", true).toBool();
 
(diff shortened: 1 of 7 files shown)


LebedevRI:
> Err, i'm not sure the feature bit is right. Let's first deal with the bug one :)

#1238.

Issue #1239, opened 2023-07-18
Consistent race condition/use-after-free in `~MusixmatchLyricsProvider`


LebedevRI:
Due to the lazy destructors, `NetworkAccessManager` may get actually fully destructed
before `MusixmatchLyricsProvider` will be (through `LyricsProviders`).

But if `MusixmatchLyricsProvider` had just sent an lyrics request,
and did not get a reply yet, it will have a `QNetworkReply*` in `replies_`,
and when `NetworkAccessManager` destructs, the object, to which that pointer points to,
is also destructed. And once `~MusixmatchLyricsProvider()` runs later,
it tries to operate on freed memory...

Locally, this very consistently (100% of the strawberry runs) results in a crash.

I've looked at this, and i don't quite understand how this is supposed to be resolved.

LebedevRI:
Actually, never mind, this appears to work: https://github.com/strawberrymusicplayer/strawberry/pull/1240
... but is rather intrusive. Is that the right fix here?

jonaski:
I think everything under Application should be shared pointers so objects are deleted in the correct order, I'm working on it.

LebedevRI:
Yes, that would be better indeed.
Thank you.

LebedevRI:
Thank you!

PR #1240, opened 2023-07-18
`Application::~Application()`: manually delete lyrics providers

LebedevRI:
This must be done before the network manager is deleted.
Due to the lazy destruction during `Lazy<>.reset()`,
this appears to be one of the approaches that solves that.

Fixes https://github.com/strawberrymusicplayer/strawberry/issues/1239

### DIFF
--- a/src/core/application.cpp
+++ b/src/core/application.cpp
@@ -233,6 +233,13 @@ Application::Application(QObject *parent)
 
 Application::~Application() {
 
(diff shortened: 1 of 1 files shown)


LebedevRI:
I'm guessing the better (proper?) solution would be to construct `LyricsProvider`'s
with a `std::shared_ptr<NetworkAccessManager>`, but that involves invasive changes to the `Lazy<>`.

PR #1246, opened 2023-07-26
Mpris2: Add new property to read/write the rating

Dakes:
This is to create shortcuts to set the rating and/or to set it from scripts.
This feature was wished for in these threads: 
https://forum.strawberrymusicplayer.org/topic/99/shortcuts-for-ratings
https://forum.strawberrymusicplayer.org/topic/326/set-ratings-per-dbus-shortcut
https://forum.strawberrymusicplayer.org/topic/879/expose-rating-of-songs-to-automation-shortcuts

This is technically out of spec for the [mpris specification](https://specifications.freedesktop.org/mpris-spec/2.2/Player_Interface.html). But since it does not modify any of the original methods it should be fine. If you prefer I could move this to a different bus name like `org.strawberry.Player.Rating` or something. 

Currently it can for example be accessed like this: `qdbus org.mpris.MediaPlayer2.strawberry /org/mpris/MediaPlayer2 org.mpris.MediaPlayer2.Player.Rating` to get the rating of the currently playing song (As float from 0-1)
Or `qdbus org.mpris.MediaPlayer2.strawberry /org/mpris/MediaPlayer2 org.mpris.MediaPlayer2.Player.Rating 0.6` to set the rating to a float, here 0.6. (from 0-1)

### DIFF
--- a/src/core/mpris2.cpp
+++ b/src/core/mpris2.cpp
@@ -212,6 +212,7 @@ void Mpris2::EmitNotification(const QString &name) {
   else if (name == "LoopStatus") value = LoopStatus();
   else if (name == "Shuffle") value = Shuffle();
   else if (name == "Metadata") value = Metadata();
+  else if (name == "Rating") value = Rating();
   else if (name == "Volume") value = Volume();
   else if (name == "Position") value = Position();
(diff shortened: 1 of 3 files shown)


jonaski:
Thanks

PR #1253, opened 2023-08-02
[WIP] Output audio queue

LebedevRI:
This is incomplete.

Currently, the queue is placed before all of our elements,
which is great to precache input and not stutter through
the times where there's IO latency spikes,
but if there's a CPU latency spike, we *may* still stutter
because we still need to process the input.

By also caching the final output, we should be safe from that.

This introduces subtleties: when changing parameters
of said modules, the changes will no longer be immediate,
but delayed by the cache length. So we must flush the pipeline's caches.

This sounds simple, but the obvious approach does not work,
and the one that works (seek to the current position)
causes temporary stutters. This isn't nice.

Also, probes need to be taken *after* the output queue.

Also, the way the fader is implemented just doesn't work with this approach.

But overall, this works better than i expected.

### DIFF
--- a/src/engine/gstenginepipeline.cpp
+++ b/src/engine/gstenginepipeline.cpp
@@ -1558,6 +1558,17 @@ bool GstEnginePipeline::Seek(const qint64 nanosec) {
 
 }
 
+static void flush_pipeline(GstElement *pipeline) {
+  // FIXME: is this really the right way to flush pipeline without seeking?
+  gint64 position_ns_;
+  gst_element_query_position(pipeline, GST_FORMAT_TIME, &position_ns_);
+  gst_element_seek_simple (pipeline, GST_FORMAT_TIME, GST_SEEK_FLAG_FLUSH, position_ns_);
+}
+
+QFuture<void> GstEnginePipeline::FlushPipeline() {
+  return QtConcurrent::run(&set_state_threadpool_, &flush_pipeline, pipeline_);
+}
+
 void GstEnginePipeline::SetEBUR128LoudnessNormalizingGain_dB(const double ebur128_loudness_normalizing_gain_db) {
 
   ebur128_loudness_normalizing_gain_db_ = ebur128_loudness_normalizing_gain_db;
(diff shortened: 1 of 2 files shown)


jonaski:
Since "use-buffering" is set to false for the input queue, low/high watermark settings are ignored,  I'm not sure about "max-size-time", but not much need for the loop just for one extra line.

LebedevRI:
> Since "use-buffering" is set to false for the input queue, low/high watermark settings are ignored, I'm not sure about "max-size-time", but not much need for the loop just for one extra line.

Ah. That's because i thought that it would only disable messages, but not buffering altogether,

I'm mainly just not sure where to start untangling all this (+gapless playback, +proper faders),
since it's all rather interconnected //and// does not have any non-manual tests.

LebedevRI:
> > Since "use-buffering" is set to false for the input queue, low/high watermark settings are ignored, I'm not sure about "max-size-time", but not much need for the loop just for one extra line.
> 
> Ah. That's because i thought that it would only disable messages, but not buffering altogether,

After looking through [gstqueue2.c](https://github.com/Kurento/gstreamer/blob/0c6afcbeb26c0d54f816413bb29bcfc9150608f4/plugins/elements/gstqueue2.c#L1032) it is not obvious to me why `"use-buffering"=false`
would disable the whole buffering and not just the status messages.

> I'm mainly just not sure where to start untangling all this (+gapless playback, +proper faders), since it's all rather interconnected //and// does not have any non-manual tests.

jonaski:
I think you are right about this, Strawberry is a bit too sensitive to CPU high load, more so than compared to for example VLC. I've experienced this a few times when I've been compiling Qt at the same time as playing music, I think it is also worse on Windows than on Linux.
But I don't understand how we can place a queue after the volumes without issues. I've tried to experiment with this too, see the gstenginepipeline_queue branch.
Maybe we could ask on the GStreamer mailinglist about this? Are you an the devel mailinglist?

jonaski:
Sorry that I've created conflicts for you now.

LebedevRI:
> I think you are right about this, Strawberry is a bit too sensitive to CPU high load, more so than compared to for example VLC. I've experienced this a few times when I've been compiling Qt at the same time as playing music, I think it is also worse on Windows than on Linux.

> But I don't understand how we can place a queue after the volumes without issues. I've tried to experiment with this too, see the gstenginepipeline_queue branch.

What kind of issues, specifically?
The PR as-is already "just" works for me, the main problem here is cache invalidation.

That being said, please maybe don't merge this/that, i believe i want to first deal with
the gap-less playback issue first. It should be easier to start untangling from that side.

> Maybe we could ask on the GStreamer mailinglist about this?

> Are you an the devel mailinglist?

No.

PR #1299, opened 2023-10-22
CMake: Fix KDSingleApplication package name for Qt5

WOnder93:
The name of the Qt5 KDSingleApplication CMake package is just "KDSingleApplication", not "KDSingleApplication-qt".


jonaski:
Thanks

PR #1307, opened 2023-11-03
Use proper plural forms for number of tracks shown in the status

Kicer86:
This PR uses Qt's mechanism to nicely use plural forms when showing number of tracks

### DIFF
--- a/cmake/Translations.cmake
+++ b/cmake/Translations.cmake
@@ -59,7 +59,7 @@ macro(add_po outfiles po_prefix)
     # Convert the .po files to .qm files
     add_custom_command(
       OUTPUT ${_qm_filepath}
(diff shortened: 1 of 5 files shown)


jonaski:
Thanks, I think GitHub was a bit unstable yesterday and the CI did not get triggered properly, I also fixed an error in the CI yesterday (c52fc90), so could you rebase this with master so the CI get's re-triggered?
Also, could you squash the last Fix typo commit with the previous commit updating pl.po.

Kicer86:
Done.

Btw what is the reason for using .pot and .po files for translations instead of Qt's .ts? 
It seems to be semi-optimal.

jonaski:
Historically I think it's for compatibility with translation services, we use Zanata now so it should support Qt translations too, in the future we might switch to Weblate which also supports it so I don't think switching would be a problem, but I'm not sure.

PR #1327, opened 2023-11-30
Correct misstyping Fesplatte

sskopnik:


jonaski:
Translations must go through Zanata: https://translate.zanata.org/iteration/view/strawberry/master

PR #1332, opened 2023-12-03
Fix crossfade crash if track decoding failed

BetterCallMolly:
If `libav` is not installed on the system and the user tries to play an ALAC file (or any other file type that requires `libav`) and has the "Cross-fade when changing track automatically" enabled, Strawberry crashes while checking if both the current and next track are on the same album.


jonaski:
Can you squash [0c70935](https://github.com/strawberrymusicplayer/strawberry/pull/1332/commits/0c7093569dce4b3345e2f678025afdf73516b945) with the previous?

jonaski (approved):

PR #1354, opened 2024-01-09
CONTRIBUTING: Clarify "Commit messages" section.

wjandrea:
Specifically:
- Clarify what "class" is referring to
- Use an actual example, which requires moving out two bits of explanation


jonaski:
Thanks, looks good, could you squash the last commit and I'll merge it.

jonaski:
It's in the document you are modifying, anyway I'll manually merge it.

PR #1361, opened 2024-01-15
Mark Subsonic MD5 authentication as recommended

cquike:
This simple MR just marks the MD5 authentication method for subsonic as recommended when editing the settings of subsonic.

### DIFF
--- a/src/settings/subsonicsettingspage.ui
+++ b/src/settings/subsonicsettingspage.ui
@@ -124,7 +124,7 @@
       <item>
        <widget class="QRadioButton" name="auth_method_md5">
         <property name="text">
-         <string>MD5 token</string>
+         <string>MD5 token (Recommended)</string>
         </property>
        </widget>
       </item>


cquike:
The reasoning for this is that the Hex authentication method is the legacy one.
From the subsonic page http://www.subsonic.org/pages/api.jsp :
"The password, either in clear text or hex-encoded with a "enc:" prefix. Since [1.13.0](http://www.subsonic.org/pages/api.jsp#versions) this should only be used for testing purposes."

jonaski (changes requested):
Please change the commit message so it's prefixed with the class.:
ie.: `SubsonicSettingsPage: Add recommended to MD5 option`

cquike:
> Please change the commit message so it's prefixed with the class.: ie.: `SubsonicSettingsPage: Add recommended to MD5 option`

Done

jonaski (approved):

Issue #1347, opened 2024-01-01
thanks for strawberry - happy 2024 - stops playback when  encountere like DRM protected wmas


Edgepath:
Hello,

thanks for strawberry fork of clementine :D 

managed to install it like this: https://dwaves.de/2023/12/09/gnu-linux-music-imho-currently-best-music-player-for-gnu-linux-and-music-creation-tools-please-donate-so-that-the-hidden-gems-of-the-internet-stay-online/

imho it's one of the best Open Source music players :+1: 

got it configured: auto play on startup

**Describe the bug**

Strawberry stops playing with error: "The stream is encrypted and decryption is not supported."

../gst/asfdemux/gstasfdemux.c(4582): gst_asf_demux_process_object (): /GstPlayBin3:pipeline-1-pipeline/GstURIDecodeBin3:uridecodebin3/GstURISourceBin:urisourcebin0/GstParseBin:parsebin0/GstASFDemux:asfdemux1

*.wma <- VLC reports: "Could not demux ASF stream: DRM protected streams are not supported."

so this is some form of DRM protected content X-D

which is fine for me if Strawberry simply skips those titles and keeps playing :)

**To Reproduce**

play some DRM protected wma. if necessary can provide the file X-D

**Expected behavior**

maybe log some error but not display error and keep playing :)

**System Information:**
 - Operating system: Debian 12 
 - Strawberry Version: Version 1.0.22

thanks!

UltimateSolar:
thanks that was FAST :D

UltimateSolar:
Hello,

how are you? (damn cold here)

got a "similar problem" "Resource not found"

../plugins/elements/gstfilesrc.c(553): gst_file_src_start (): /GstPlayBin3:pipeline-5-pipeline/GstURIDecodeBin3:uridecodebin3/GstURISourceBin:urisourcebin12/GstFileSrc:filesrc12: No such file "/home/user/Music/Dream Theater/dream theater/Dream Theater - Fatal Trage.mp3"

<- the file was deleted because too much tragedy is no good, strawberry halts playback. why? just go to next title :D

just keep em' play'n bro :D

PR #1369, opened 2024-02-03
Playlist: skip invalid songs to prevent playback interruptions

JcKlomp:
Also skips invalid songs if the option to gray out unavailable songs is enabled.

When the option to gray out unavailable songs on playback is enabled, this will now also skip the song after they become invalid.
Once invalid they will be skipped until they are valid again, which requires manually selecting the song.
Before songs become invalid there still is an error message and playback gets interrupted.

When the option to gray out unavailable song on startup is enabled, all invalid grayed out songs are skipped during playback.
This means that playback can be expected to be uninterrupted, even when there is a great amount of missing songs in the playlist for whatever reason.

This simply uses the skip function to skip/unskip invalid songs.
Side effect is that the visual appearance of grayed out songs changes so these will also have a strikethrough formatting.
Also if a manually skipped song becomes invalid, then gets fixed and becomes valid again it will no longer be skipped.
However since user action is required for this it could be considered expected behaviour.

This change is intentionally kept minimal initially, since it is not completely clear to me whether this is the right approach.
There are no changes to surrounding code or comments.
Nor are the descriptions changed of the 2 options for graying out song.
Renaming or rephrasing the options, or introducing an additional option might be more confusing than just changing the behaviour.
If necessary possibly best would be to only rephrase the 2 options to 'gray out and skip unavailable song' to convey the behaviour.

To me, also skipping unavailable songs is more expected than only graying them out when the option is enabled, and then still attempting to play them which is likely to fail.
Especially on large playlists that gets shared between different systems (but where not all of the files are synced too) this can lead to repeated unnecessary playback interruptions.
The function to remove unavailable song from the playlist is a nice workaround, but it is a bit hidden and requires user action.
Plus having the unavailable songs still visible in the playlist can be handy when wanting to resolve the issue with the files at a later time without having to mess with different versions of the playlist.

Vaguely related to #1347

### DIFF
--- a/src/playlist/playlist.cpp
+++ b/src/playlist/playlist.cpp
@@ -2164,10 +2164,14 @@ void Playlist::InvalidateDeletedSongs() {
         // gray out the song if it's not there
         item->SetForegroundColor(kInvalidSongPriority, kInvalidSongColor);
         invalidated_rows.append(row);  // clazy:exclude=reserve-candidates
+        // skip invalid song
+        item->SetShouldSkip(true);
       }
       else if (exists && item->HasForegroundColor(kInvalidSongPriority)) {
         item->RemoveForegroundColor(kInvalidSongPriority);
         invalidated_rows.append(row);  // clazy:exclude=reserve-candidates
+        // song is valid again; do not skip it
+        item->SetShouldSkip(false);
       }
     }
   }
@@ -2285,9 +2289,13 @@ bool Playlist::ApplyValidityOnCurrentSong(const QUrl &url, const bool valid) {
     // Gray out the song if it's now broken; otherwise undo the gray color
     if (valid) {
       current->RemoveForegroundColor(kInvalidSongPriority);
+      // Song is valid again; do not skip it
+      current->SetShouldSkip(false);
     }
     else {
       current->SetForegroundColor(kInvalidSongPriority, kInvalidSongColor);
+      // Skip invalid song
+      current->SetShouldSkip(true);
     }
   }
 


jonaski:
I'm not sure I like this change. One example is when songs are grayed because the collection directory is unavailable, it requires a full scan to get them back, if you it skips the song to it's impossible to retry the songs without having to wait for the scan to complete.

JcKlomp:
Those could still be re-enabled by manually playing or unskipping the songs?
(I don't think I fully grasp the situation there.)

But I checked all the options a bit better and turns out what I need is already largely available in the form of the option `continue_on_error`.
I guess I had that enabled only briefly while trying out these options and loading a playlist at runtime, which doesn't trigger the `greyout_songs_startup` function.
So at that point the unavailable songs were not yet known which confused me when cleaning the playlist.

There still could be many repeated error messages but at least playback is not interrupted.
To me not as ideal as simply skipping songs which are _known_ to not be available, but I agree it might not always be expected or desired behaviour.

Adding an additional option for an auto skip option that supplements the greyout option is pretty straightforward too (and I have a _rough_ version running with that), but it does not sound too appealing to me because that settings page already looks a bit complex.
If visually grouped together with the greyout options it may still look clear enough.

But perhaps only adding a user-selectable function to the playlist manager to skip all unavailable songs would be simpler and more acceptable?
(Preferably combined with making `greyout_songs_startup` also take effect for loading playlists at runtime.)

I think I'll try to have another look at this at a later time.

jonaski:
> Those could still be re-enabled by manually playing or unskipping the songs? (I don't think I fully grasp the situation there.)
> 
Yes they can, but I still do not like the behavior, but if you really want it, please make it optional as a checkbox in the playlist settings.

JcKlomp:
Okay, I'm going to look at cleaning up my code for making this configurable.
For now I'll close this because at the moment it is not sufficient and I don't want to rush it while keeping this PR lingering.
Thank you for looking into this.

PR #1371, opened 2024-02-04
Display song progress on taskbar

buckmelanoma:
This PR adds a new function to display the song progress on the OS taskbar.  

image

image

image

The function has three parameters:
- visible: toggles whether progress is visible or hidden on the taskbar
- position: the position of playback within the current song
- length: length of the current song

Test plan:
- Tested with local files and internet streams (taskbar progress hidden with streams)
- Tested skipping to previous and next tracks
- Tested opening and closing player with "resume playback" enabled and disabled
- Tested on top Linux desktop environments - KDE Plasma 5 and 6, GNOME
- Untested with Windows and macOS (I lack access to both)

### DIFF
--- a/src/core/mainwindow.cpp
+++ b/src/core/mainwindow.cpp
@@ -73,8 +73,10 @@
 #include <QToolButton>
 #include <QCheckBox>
 #include <QClipboard>
+#ifdef HAVE_DBUS
 #include <QDBusConnection>
 #include <QDBusMessage>
+#endif
 
 #include "core/logging.h"
 
@@ -1272,7 +1274,9 @@ void MainWindow::Exit() {
       }
     }
 
+#ifdef HAVE_DBUS
     UpdateTaskbarProgress(false, 0, 0);
+#endif
 
     DoExit();
   }
@@ -1326,7 +1330,9 @@ void MainWindow::MediaStopped() {
   tray_icon_->SetProgress(0);
   tray_icon_->SetStopped();
 
+#ifdef HAVE_DBUS
   UpdateTaskbarProgress(false, 0, 0);
+#endif
 
   song_playing_ = Song();
   song_ = Song();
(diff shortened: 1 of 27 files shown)


src/core/mainwindow.cpp
```diff
@@ -1726,6 +1739,21 @@ void MainWindow::UpdateTrackSliderPosition() {
 
 }
 
+void MainWindow::UpdateTaskbarProgress(const bool visible, const double position, const double length) {
+
+  QVariantMap map;
+  QDBusMessage msg = QDBusMessage::createSignal(QStringLiteral("/org/strawberrymusicplayer/strawberry"),
+		                                QStringLiteral("com.canonical.Unity.LauncherEntry"),
```
jonaski, line 1746: Please avoid the use of tabs instead of spaces. Here is also a trailing space at the end of the line.
But I prefer to keep this on one line.

src/core/mainwindow.cpp
```diff
@@ -73,6 +73,8 @@
 #include <QToolButton>
 #include <QCheckBox>
 #include <QClipboard>
+#include <QDBusConnection>
```
jonaski, line 76: You need to use `#ifdef HAVE_DBUS` around all this code including the function.

jonaski (changes requested):

src/core/mainwindow.cpp
buckmelanoma: Thanks, that's a good call.  I'll make the changes and update the PR.

src/core/mainwindow.cpp
buckmelanoma: Done

src/core/mainwindow.cpp
buckmelanoma: Done

jonaski:
Merged dd6b9bb38d1ffe620ee938b44aefefc9e7938f38

Issue #1287, opened 2023-10-10
Option to save lyrics into tags


guidomz:
Some lyrics related options:
* "Save to tags" button in the lyrics display panel
* a setting to "Automatically save lyrics as tags if not yet there" or something
* an option to automatically save lyrics to tags when scanning audio files into library

jonaski:
Is this something you plan to do?
We shouldn't save anything during scanning, it has to be a separate function.

guidomz:
Sorry, I don't know coding in C. Those options are actually just suggestions for improvement, which I should have mentioned. I agree not saving during scanning.

jonaski:
As stated in the README we can't take any requests now as I'm pretty much the only contributor, and I already have things that are still not finished after 2 years. Having issues open for suggestions/features on GitHub would potentially result in hundreds of issues, impossible to do, therefore it's strictly bugs for now. It might make sense if someone has already started to work on something, so we don't do the same things, but other than that it's pointless without more contributors.

PR #1391, opened 2024-03-06
EditTagDialog: Add a button to fetch lyrics in tags edition dialog

Perdu:
Strawberry is an awesome player having the ability to both fetch lyrics and edit tags. This PR aims at fetching lyrics directly into tags a bit faster (as required e.g. in #1287), by adding a "Complete lyrics automatically" in the lyrics tab of the tags edition dialog.
 
It probably needs a bit more polishing. I'd be happy to get help on a few points:
- I'm not sure `data_[ui_->song_list->selectionModel()->selectedIndexes().first().row()].original_;` was the most straightforward way to get the current song, but I didn't find another way
- I had to use `Q_UNUSED` twice because two of the parameters of the signal emitted by the LyricsFetcher are not useful here (not sure how to make a signal without these)
- might be better to display a loading icon and somehow display an error when nothing is found, but I don't know how to do both (adding "not found" in text area doesn't seem great)

### DIFF
--- a/src/context/contextview.cpp
+++ b/src/context/contextview.cpp
@@ -631,7 +631,11 @@ void ContextView::UpdateLyrics(const quint64 id, const QString &provider, const 
 
   if (static_cast<qint64>(id) != lyrics_id_) return;
 
-  lyrics_ = lyrics + "\n\n(Lyrics from " + provider + ")\n";
+  if (lyrics == nullptr) {
+    lyrics_ = "\n\nLyrics not found.\n";
+  } else {
+    lyrics_ = lyrics + "\n\n(Lyrics from " + provider + ")\n";
+  }
   lyrics_id_ = -1;
 
   if (action_show_lyrics_->isChecked() && !lyrics_.isEmpty()) {
--- a/src/context/contextview.cpp
+++ b/src/context/contextview.cpp
@@ -631,7 +631,7 @@ void ContextView::UpdateLyrics(const quint64 id, const QString &provider, const 
 
   if (static_cast<qint64>(id) != lyrics_id_) return;
 
-  if (lyrics == nullptr) {
+  if (lyrics.isEmpty()) {
     lyrics_ = "\n\nLyrics not found.\n";
   } else {
     lyrics_ = lyrics + "\n\n(Lyrics from " + provider + ")\n";
--- a/src/context/contextview.cpp
+++ b/src/context/contextview.cpp
@@ -633,7 +633,8 @@ void ContextView::UpdateLyrics(const quint64 id, const QString &provider, const 
 
   if (lyrics.isEmpty()) {
     lyrics_ = "\n\nLyrics not found.\n";
-  } else {
+  }
+  else {
     lyrics_ = lyrics + "\n\n(Lyrics from " + provider + ")\n";
   }
   lyrics_id_ = -1;
--- a/src/dialogs/edittagdialog.cpp
+++ b/src/dialogs/edittagdialog.cpp
@@ -1421,7 +1421,7 @@ void EditTagDialog::FetchTagSongChosen(const Song &original_song, const Song &ne
 void EditTagDialog::FetchLyrics() {
 
   if (ui_->song_list->selectionModel()->selectedIndexes().isEmpty()) return;
-  const Song song = data_[ui_->song_list->selectionModel()->selectedIndexes().first().row()].current_;
+  const Song &song = data_[ui_->song_list->selectionModel()->selectedIndexes().first().row()].current_;
   lyrics_fetcher_->Clear();
   ui_->lyrics->setPlainText(tr("loading..."));
   lyrics_id_ = static_cast<qint64>(lyrics_fetcher_->Search(song.effective_albumartist(), song.artist(), song.album(), song.title()));
--- a/src/dialogs/edittagdialog.cpp
+++ b/src/dialogs/edittagdialog.cpp
@@ -1436,7 +1436,8 @@ void EditTagDialog::UpdateLyrics(const quint64 id, const QString &provider, cons
   lyrics_id_ = -1;
   if (lyrics != nullptr) {
       ui_->lyrics->setPlainText(lyrics);
-  } else {
+  }
+  else {
       ui_->lyrics->setPlainText(tr("Not found."));
   }
 }
--- a/src/dialogs/edittagdialog.cpp
+++ b/src/dialogs/edittagdialog.cpp
@@ -1423,6 +1423,7 @@ void EditTagDialog::FetchLyrics() {
   if (ui_->song_list->selectionModel()->selectedIndexes().isEmpty()) return;
   const Song song = data_[ui_->song_list->selectionModel()->selectedIndexes().first().row()].current_;
   lyrics_fetcher_->Clear();
+  ui_->lyrics->setPlainText(tr("loading..."));
   lyrics_id_ = static_cast<qint64>(lyrics_fetcher_->Search(song.effective_albumartist(), song.artist(), song.album(), song.title()));
 
 }
@@ -1433,8 +1434,11 @@ void EditTagDialog::UpdateLyrics(const quint64 id, const QString &provider, cons
 
   if (static_cast<qint64>(id) != lyrics_id_) return;
   lyrics_id_ = -1;
-  ui_->lyrics->setPlainText(lyrics);
-
+  if (lyrics != nullptr) {
+      ui_->lyrics->setPlainText(lyrics);
+  } else {
+      ui_->lyrics->setPlainText(tr("Not found."));
+  }
 }
 
 void EditTagDialog::SongSaveTagsComplete(TagReaderReply *reply, const QString &filename, Song song, const UpdateCoverAction cover_action) {
--- a/src/dialogs/edittagdialog.cpp
+++ b/src/dialogs/edittagdialog.cpp
@@ -120,7 +120,8 @@ EditTagDialog::EditTagDialog(Application *app, QWidget *parent)
       summary_cover_art_id_(-1),
       tags_cover_art_id_(-1),
       cover_art_is_set_(false),
-      save_tag_pending_(0) {
+      save_tag_pending_(0),
+      lyrics_id_(-1) {
 
   QObject::connect(&*app_->album_cover_loader(), &AlbumCoverLoader::AlbumCoverLoaded, this, &EditTagDialog::AlbumCoverLoaded);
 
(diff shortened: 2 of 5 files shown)


src/dialogs/edittagdialog.cpp
```diff
@@ -1411,6 +1415,23 @@ void EditTagDialog::FetchTagSongChosen(const Song &original_song, const Song &ne
 
 }
 
+void EditTagDialog::FetchLyrics() {
+
+  Song song = data_[ui_->song_list->selectionModel()->selectedIndexes().first().row()].original_;
```
jonaski, line 1420: This can be const reference.

Also, you need to check that `selectedIndexes` is not empty, otherwise this will crash.

Add `if (ui_->song_list->selectionModel()->selectedIndexes().isEmpty()) return;` before this line.

src/dialogs/edittagdialog.cpp
```diff
@@ -1411,6 +1415,23 @@ void EditTagDialog::FetchTagSongChosen(const Song &original_song, const Song &ne
...
+  lyrics_fetcher_->Clear();
+  lyrics_fetcher_->Search(song.effective_albumartist(), song.artist(), song.album(), song.title());
```
jonaski, line 1422: You need to save the ID to a member variable, so if the user selects another song while the lyrics are fetched, it does not use that lyrics.

src/dialogs/edittagdialog.cpp
```diff
@@ -1411,6 +1415,23 @@ void EditTagDialog::FetchTagSongChosen(const Song &original_song, const Song &ne
...
+
+}
+
+void EditTagDialog::UpdateLyrics(const quint64 id, const QString &provider, const QString &lyrics) {
+
+  Q_UNUSED(id);
```
jonaski, line 1428: You need to check that the ID here matched the one returned by Search().

jonaski (changes requested):

src/dialogs/edittagdialog.cpp
jonaski: I also think `current_` should be used here, not `original_`, so if the user changes the tags it uses the modified metadata.

Perdu:
Comments addressed.
I also attempted to add messages:
- A "loading" message is simple (`ui_->lyrics->setPlainText(tr("loading..."));` at edittagdialog.cpp:1426
- "Not found" is not straightforward. When no lyrics are found, no signal is emitted (lyricsfetchersearch.cpp:135). I could add one with empty lyrics and provider, but that would probably have to be handled in other places to avoid crashes

Perdu:
Never mind, there was actually only one extra case to handle. I added an error message for that case too.

src/context/contextview.cpp
```diff
@@ -631,7 +631,11 @@ void ContextView::UpdateLyrics(const quint64 id, const QString &provider, const
 
   if (static_cast<qint64>(id) != lyrics_id_) return;
 
-  lyrics_ = lyrics + "\n\n(Lyrics from " + provider + ")\n";
+  if (lyrics == nullptr) {
```
jonaski, line 634:
```suggestion
  if (lyrics.isEmpty()) {
```

src/context/contextview.cpp
```diff
@@ -631,7 +631,11 @@ void ContextView::UpdateLyrics(const quint64 id, const QString &provider, const
...
+    lyrics_ = "\n\nLyrics not found.\n";
+  } else {
```
jonaski, line 636:
```suggestion
  }
  else {
```

src/dialogs/edittagdialog.cpp
```diff
@@ -112,13 +113,15 @@ EditTagDialog::EditTagDialog(Application *app, QWidget *parent)
...
+      lyrics_fetcher_(new LyricsFetcher(app->lyrics_providers(), this)),
       image_no_cover_thumbnail_(ImageUtils::GenerateNoCoverImage(QSize(128, 128), devicePixelRatioF())),
       loading_(false),
       ignore_edits_(false),
       summary_cover_art_id_(-1),
       tags_cover_art_id_(-1),
       cover_art_is_set_(false),
-      save_tag_pending_(0) {
+      save_tag_pending_(0),
+      lyrics_id_(-1) {
```
jonaski, line 124:
```suggestion
      lyrics_fetch_id_(-1) {
```

src/lyrics/lyricsfetchersearch.cpp
```diff
@@ -133,6 +133,8 @@ void LyricsFetcherSearch::AllProvidersFinished() {
   if (!results_.isEmpty()) {
     qLog(Debug) << "Using lyrics from" << results_.last().provider << "for" << request_.artist << request_.title << "with score" << results_.last().score;
     emit LyricsFetched(id_, results_.last().provider, results_.last().lyrics);
+  } else {
+    emit LyricsFetched(id_, nullptr, nullptr);
```
jonaski, line 137:
```suggestion
    emit LyricsFetched(id_, QString(), QString());
```

src/dialogs/edittagdialog.cpp
```diff
@@ -1411,6 +1418,29 @@ void EditTagDialog::FetchTagSongChosen(const Song &original_song, const Song &ne
 
 }
 
+void EditTagDialog::FetchLyrics() {
+
+  if (ui_->song_list->selectionModel()->selectedIndexes().isEmpty()) return;
+  const Song song = data_[ui_->song_list->selectionModel()->selectedIndexes().first().row()].current_;
```
jonaski, line 1424:
```suggestion
  const Song &song = data_[ui_->song_list->selectionModel()->selectedIndexes().first().row()].current_;
```

src/dialogs/edittagdialog.cpp
```diff
@@ -1411,6 +1418,29 @@ void EditTagDialog::FetchTagSongChosen(const Song &original_song, const Song &ne
...
+
+}
+
+void EditTagDialog::UpdateLyrics(const quint64 id, const QString &provider, const QString &lyrics) {
+
+  Q_UNUSED(provider);
+
+  if (static_cast<qint64>(id) != lyrics_id_) return;
+  lyrics_id_ = -1;
+  if (lyrics != nullptr) {
```
jonaski, line 1437:
```suggestion
  if (lyrics.isEmpty()) {
```

`lyrics.isEmpty()` statement first, and lyrics found in `else` so it's simiar to `ContextView`.

src/dialogs/edittagdialog.cpp
```diff
@@ -1411,6 +1418,29 @@ void EditTagDialog::FetchTagSongChosen(const Song &original_song, const Song &ne
...
+      ui_->lyrics->setPlainText(lyrics);
+  } else {
```
jonaski, line 1439:
```suggestion
  }
  else {
```

jonaski (changes requested):

src/lyrics/lyricsfetchersearch.cpp
jonaski: `lyrics` is not a pointer, but a reference so `nullptr` should not be used, `QString::isEmpty()` covers both empty null strings.

src/context/contextview.cpp
jonaski: `lyrics` is not a pointer, but a reference so `nullptr` should not be used, `QString::isEmpty()` covers both empty and null strings.

src/dialogs/edittagdialog.cpp
Perdu: I copied that name from context/contexview.cpp, in which a similar variable does the same thing. I think it should remain the same for consistency (or be changed in both)

src/dialogs/edittagdialog.cpp
Perdu: case got inverted

Perdu: Fixed in separate commit

src/dialogs/edittagdialog.cpp
```diff
@@ -1411,6 +1418,30 @@ void EditTagDialog::FetchTagSongChosen(const Song &original_song, const Song &ne
...
+}
+
+void EditTagDialog::UpdateLyrics(const quint64 id, const QString &provider, const QString &lyrics) {
+
+  Q_UNUSED(provider);
+
+  if (static_cast<qint64>(id) != lyrics_id_) return;
+  lyrics_id_ = -1;
+  if (!lyrics.isEmpty()) {
+      ui_->lyrics->setPlainText(lyrics);
```
jonaski, line 1438: indent should be 2 spaces, not 4.

src/dialogs/edittagdialog.cpp
```diff
@@ -1411,6 +1418,30 @@ void EditTagDialog::FetchTagSongChosen(const Song &original_song, const Song &ne
...
+
+}
+
+void EditTagDialog::UpdateLyrics(const quint64 id, const QString &provider, const QString &lyrics) {
+
+  Q_UNUSED(provider);
+
+  if (static_cast<qint64>(id) != lyrics_id_) return;
+  lyrics_id_ = -1;
+  if (!lyrics.isEmpty()) {
```
jonaski, line 1437:
```suggestion
  if (lyrics.isEmpty()) {
```
Please move the not found section to the `if`, and setting lyrics to `else`.

src/dialogs/edittagdialog.cpp
```diff
@@ -1411,6 +1418,30 @@ void EditTagDialog::FetchTagSongChosen(const Song &original_song, const Song &ne
...
+      ui_->lyrics->setPlainText(lyrics);
+  }
+  else {
+      ui_->lyrics->setPlainText(tr("Not found."));
```
jonaski, line 1441: indent should be 2 spaces, not 4.

src/dialogs/edittagdialog.cpp
```diff
@@ -1411,6 +1418,30 @@ void EditTagDialog::FetchTagSongChosen(const Song &original_song, const Song &ne
...
+  }
+}
```
jonaski, line 1443: Newline before `}` so it's consistent with the rest of the codebase.

src/lyrics/lyricsfetchersearch.cpp
```diff
@@ -133,6 +133,8 @@ void LyricsFetcherSearch::AllProvidersFinished() {
   if (!results_.isEmpty()) {
     qLog(Debug) << "Using lyrics from" << results_.last().provider << "for" << request_.artist << request_.title << "with score" << results_.last().score;
     emit LyricsFetched(id_, results_.last().provider, results_.last().lyrics);
+  } else {
```
jonaski, line 136:
```suggestion
  }
  else {
```

jonaski (changes requested):
Looks good except for minor formatting, please fix and squash the commits and I'll merge this.

We should implement lyrics fetching similar to what is done for missing tags where the user can select from available provider / lyrics which one to use, but we can do that later.

jonaski:
Thanks, merged https://github.com/strawberrymusicplayer/strawberry/commit/27ee6e764323553437ce5dd2a81fa45ad8e4a90c

PR #1393, opened 2024-03-09
Add an option not to skip “The” when sorting artist names

sboukortt:


src/internet/internetsearchmodel.cpp
```diff
@@ -95,7 +95,7 @@ QStandardItem *InternetSearchModel::BuildContainers(const Song &s, QStandardItem
       }
       else {
         display_text = CollectionModel::TextOrUnknown(s.effective_albumartist());
-        sort_text = CollectionModel::SortTextForArtist(s.effective_albumartist());
+        sort_text = CollectionModel::SortTextForArtist(s.effective_albumartist(), /*skip_articles=*/true);
```
jonaski, line 98: Please remove the comments

src/settings/collectionsettingspage.ui
```diff
@@ -2,14 +2,6 @@
 <ui version="4.0">
  <class>CollectionSettingsPage</class>
  <widget class="QWidget" name="CollectionSettingsPage">
-  <property name="geometry">
```
jonaski, line 5: Any idea why this is removed here?

src/settings/collectionsettingspage.ui
```diff
@@ -225,6 +217,13 @@ If there are no matches then it will use the largest image in the directory.</st
         </property>
        </widget>
       </item>
+      <item>
+       <widget class="QCheckBox" name="sort_skips_articles">
+        <property name="text">
+         <string>Skip articles such as &quot;The&quot; when sorting artist names</string>
```
jonaski, line 223: such as "the", "a" and "an"

src/collection/collectionmodel.h
```diff
@@ -173,7 +176,7 @@ class CollectionModel : public SimpleTreeModel<CollectionItem> {
   static QString PrettyDisc(const int disc);
   static QString SortText(QString text);
   static QString SortTextForNumber(const int number);
-  static QString SortTextForArtist(QString artist);
+  static QString SortTextForArtist(QString artist, bool skip_articles);
```
jonaski, line 179: const bool skip_articles

src/collection/collectionmodel.cpp
```diff
@@ -1719,15 +1727,17 @@ QString CollectionModel::SortText(QString text) {
 
 }
 
-QString CollectionModel::SortTextForArtist(QString artist) {
+QString CollectionModel::SortTextForArtist(QString artist, bool skip_articles) {
```
jonaski, line 1730: const bool skip_articles

jonaski (changes requested):

src/internet/internetsearchmodel.cpp
sboukortt: Done (the reason for having them was https://abseil.io/tips/94).

src/settings/collectionsettingspage.ui
sboukortt: Is it fine if I drop “such as” if it’s going to be an exhaustive list? Or might the list be expanded later (for example to articles in other languages)?

src/collection/collectionmodel.h
sboukortt: Done, but note that in declarations (as opposed to definitions), it makes no difference:

https://abseil.io/tips/109

> Never use top-level `const` on function parameters in _declarations_ that are not definitions (and be careful not to copy/paste a meaningless `const`). It is meaningless and ignored by the compiler, it is visual noise, and it could mislead readers.

src/collection/collectionmodel.cpp
sboukortt: Done, thanks.

src/settings/collectionsettingspage.ui
jonaski: Since it's only 3 for now let's skip "such as".

jonaski (approved):

jonaski:
Thanks

PR #1452, opened 2024-06-03
Add volume increment setting

jjmith:


src/core/player.cpp
```diff
@@ -700,16 +702,15 @@ void Player::SetVolume(const uint volume) {
...
   uint old_volume = GetVolume();
-  uint new_volume = std::min(old_volume + 5, static_cast<uint>(100));
+  uint new_volume = std::min(old_volume + volume_increment_, static_cast<uint>(100));
   if (new_volume == old_volume) return;
   SetVolume(new_volume);
 
 }
 
 void Player::VolumeDown() {
-
```
jonaski, line 710: nitpick: don't remove newline here.

jonaski (changes requested):
Thanks, could you add back the removed newline and squash the new commit and I'll merge this.

jjmith:
Thanks! I think it should be fixed now.

jonaski (approved):

PR #1457, opened 2024-06-07
Fix integer overflow in free space bar

whatwareweb:
This fixes a bug in the free space bar. If more songs were added to a device than the device had storage capacity for, there would be an integer underflow and it would show that the device had an incorrect amount of storage available as in the picture.
image
This fix checks for that condition and replaces it as shown in this picture
image

### DIFF
--- a/src/widgets/freespacebar.cpp
+++ b/src/widgets/freespacebar.cpp
@@ -66,6 +66,7 @@ FreeSpaceBar::FreeSpaceBar(QWidget *parent)
       total_(100),
       free_text_(tr("Available")),
       additional_text_(tr("New songs")),
+      exceeded_text_(tr("Exceeded by")),
       used_text_(tr("Used")) {
 
   setMinimumHeight(FreeSpaceBar::sizeHint().height());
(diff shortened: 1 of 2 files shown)


src/widgets/freespacebar.cpp
```diff
@@ -194,7 +195,11 @@ void FreeSpaceBar::DrawText(QPainter *p, const QRect r) {
   if (additional_ > 0) {
     labels << Label(TextForSize(additional_text_, additional_), kColorAdd1);
   }
-  labels << Label(TextForSize(free_text_, free_ - additional_), kColorBg2);
+  if (free_ > additional_ || additional_ == 0) {
+    labels << Label(TextForSize(free_text_, free_ - additional_), kColorBg2);
+  } else {
```
jonaski, line 200:
```suggestion
  }
  else {
```

Please follow same code style as the rest, please correct and squash commit.

jonaski (changes requested):
Thanks for the fix, minor code style issue, otherwise good. Please squash correction commits with the first.

whatwareweb:
OK it's fixed.

jonaski (approved):

PR #1458, opened 2024-06-08
Fix GST_FRAMES_TO_CLOCK_TIME frames parameter

guprobr:
Perphaps divide frames per channels instead of multiply, since its a duration...
This fixes truncated visual output of 24 bit audio in analyzers that do not apply any transformation

### DIFF
--- a/src/engine/gstenginepipeline.cpp
+++ b/src/engine/gstenginepipeline.cpp
@@ -1140,7 +1140,7 @@ GstPadProbeReturn GstEnginePipeline::BufferProbeCallback(GstPad *pad, GstPadProb
     }
     gst_buffer_unmap(buf, &map_info);
     buf16 = gst_buffer_new_wrapped(d, buf16_size);
-    GST_BUFFER_DURATION(buf16) = GST_FRAMES_TO_CLOCK_TIME(samples * sizeof(int16_t) * channels, rate);
+    GST_BUFFER_DURATION(buf16) = GST_FRAMES_TO_CLOCK_TIME(samples * sizeof(int16_t) / channels, rate);
     buf = buf16;
 
     instance->logged_unsupported_analyzer_format_ = false;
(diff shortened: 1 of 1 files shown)


jonaski (changes requested):
Can you please fix edit the commit message `git commit --amend`, and change the first line to something like.:

`GstEnginePipeline: Divide samples and format by channels for buffer duration`

This way it's clearer what class is changed when you look at a commit message.

guprobr:
of course! just a sec

guprobr:
I'm kinda of lame in terms of git, rebase... is this correct?

pick 45bad3be Fix integer underflow bug
squash 0914f79c GstEnginePipeline: Divide samples and format by channels for buffer duration

jonaski:
To fix the commit message you can just use: `git commit --amend`,
or with rebase you can do.: `git rebase -i 45bad3be04858aec880861e94fb9b8cf6edc3cd0~`
Where `45bad3be04858aec880861e94fb9b8cf6edc3cd0` is your commit.
Then you enter `r` in front of the commit instead of `pick`.
And to push the changes.:
`git push origin channels-at-gst_frames_to_clock_time-fix --force-with-lease`
No need to squash since there is just one commit, removed my comment about that.

jonaski:
I have now tested this fix with S16LE, S24LE, S32LE and F32LE, it looks good. There is just one format I didn't test and that's S24_32LE, I don't remember how I got that format.

guprobr:
S24_32LE:

**S**: Signed integer - This indicates the audio data uses negative values to represent silence or quieter sounds and positive values for louder sounds.
**24**: 24 bits per sample - This refers to the number of bits used to represent the amplitude of a single audio sample. In this case, 24 bits provide a higher resolution compared to formats like 16-bit audio.
**LE**: Little-endian byte order - This specifies the order bytes are arranged within a multi-byte value. In little-endian, the least significant byte comes first.

 (24-bit samples stored in a 32-bit container) -> convert like this: `ffmpeg -i song.wav -c:a pcm_s24le -sample_fmt s32 output_file.wav`

jonaski (approved):

PR #1467, opened 2024-06-21
Add visualisations

jonaski:
Currently works with Qt 5 using `QGLWidget`, but that class is deprecated and needs to be replaced with `QOpenGLWidget` to work with both Qt 5 and 6, which currently crashes.


kblaschke:
Qt6 uses EGL as the default OpenGL implementation on Linux, so I fear unless we add support for building against libEGL (see issue https://github.com/projectM-visualizer/projectm/issues/681) or at least have Qt resolve the GL functions in libprojectM, this won't really work well. FBO rendering support is already in master, and will be released in v4.2, so no native surface (e.g. QOpenGlWindow) is required anymore, making things a bit easier.

PR #1494, opened 2024-07-16
CMake: Use result of find_program instead of calling xgettext directly

ajtribick:
`find_program` may be able to find the executable in locations which require a path, so use the result variable in the command.


jonaski:
Using the variable doesn't work on Windows with NMake, that's why I removed use of the variable in 9b1035a5f25b7684c84d6d8e7a5d5305b1082b3c

jonaski (approved):
I don't see any issue on Windows anymore.

PR #1544, opened 2024-09-13
Settings: Add "Restart or previous" global shortcut

abika:
Small new feature as the title says: A new global shortcut "Restart or previous" is added that has the same functionality as the "--restart-or-previous" command line argument (and uses the same method `Player::RestartOrPrevious()`).

### DIFF
--- a/src/translations/translations.pot
+++ b/src/translations/translations.pot
@@ -2000,6 +2000,9 @@ msgstr ""
 msgid "Previous track"
 msgstr ""
 
+msgid "Restart or previous track"
+msgstr ""
+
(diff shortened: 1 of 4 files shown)


src/translations/translations.pot
```diff
@@ -2000,6 +2000,9 @@ msgstr ""
 msgid "Previous track"
 msgstr ""
 
+msgid "Restart or previous track"
```
abika, line 2003: This was added somehow automatically. Not sure if it should be part of this MR?

jonaski (approved):

PR #1595, opened 2024-11-04
CMakeLists.txt: Make QtSparkle optional on Windows builds

ajtribick:
Allow proceeding with the build without setting `HAVE_QTSPARKLE` if QtSparkle is not found at configuration time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -276,7 +276,7 @@ endif()
 
 if(WIN32)
   find_package(getopt-win REQUIRED)
-  pkg_check_modules(QTSPARKLE REQUIRED IMPORTED_TARGET qtsparkle-qt${QT_VERSION_MAJOR})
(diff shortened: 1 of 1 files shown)


jonaski:
I want CMake to fail if QtSparkle isn't found so the issue can be fixed, so this would have to be a CMake option to turn specifically turn QtSparkle off, but what's the use-case for this?

ajtribick:
I'm building locally with dependencies provided by vcpkg. Currently it doesn't provide QtSparkle. Since I don't really need the functionality it provides, and most of the project makes it look like it's intended to be optional (e.g. the test for `QTSPARKLE_FOUND` immediately after the call to `pkg_check_modules`, and the various preprocessor macros) I'd like to be able to build without it.

(The other issue I ran into with vcpkg when building the current master branch is the name of the getopt target it provides, but that feels more like a vcpkg issue to me so didn't raise a change here.)

PR #1636, opened 2025-01-04
Subsonic: Add support to retrieve album covers using album id

Leohige:
I started using strawberry client exclusively for the subsonic support.

Im always updating my library, and this became a problem because it always retrieves all the covers based on the "coverArt" value, which is always different due to each of my songs having embedded covers, so if I have 2000 songs it fetches the 2000 covers. I think a lot of people struggle with that. 

Since the subsonic API supports retrieving the cover art by the album id, I've just added support for that.

image

This is optional and it can be enabled or disabled in subsonic settings page.

image

With "Use album id for album covers" option enabled:

image

With "Use album id for album covers" option disabled:

image

### DIFF
--- a/src/settings/subsonicsettingspage.cpp
+++ b/src/settings/subsonicsettingspage.cpp
@@ -51,6 +51,7 @@ SubsonicSettingsPage::SubsonicSettingsPage(SettingsDialog *dialog, const SharedP
 
   QObject::connect(ui_->button_test, &QPushButton::clicked, this, &SubsonicSettingsPage::TestClicked);
   QObject::connect(ui_->button_deletesongs, &QPushButton::clicked, &*service_, &SubsonicService::DeleteSongs);
+  QObject::connect(ui_->checkbox_download_album_covers, &QCheckBox::toggled, this, &SubsonicSettingsPage::CheckboxDownloadAlbumCoversToggled);
 
   QObject::connect(this, &SubsonicSettingsPage::Test, &*service_, &SubsonicService::SendPingWithCredentials);
 
@@ -78,6 +79,7 @@ void SubsonicSettingsPage::Load() {
   ui_->checkbox_http2->setChecked(s.value(kHTTP2, false).toBool());
   ui_->checkbox_verify_certificate->setChecked(s.value(kVerifyCertificate, false).toBool());
   ui_->checkbox_download_album_covers->setChecked(s.value(kDownloadAlbumCovers, true).toBool());
+  ui_->checkbox_use_album_id_for_album_covers->setChecked(s.value(kUseAlbumIdForAlbumCovers, false).toBool());
   ui_->checkbox_server_scrobbling->setChecked(s.value(kServerSideScrobbling, false).toBool());
 
   const AuthMethod auth_method = static_cast<AuthMethod>(s.value(kAuthMethod, static_cast<int>(AuthMethod::MD5)).toInt());
@@ -89,6 +91,8 @@ void SubsonicSettingsPage::Load() {
       ui_->auth_method_md5->setChecked(true);
       break;
   }
+
+  ui_->checkbox_use_album_id_for_album_covers->setEnabled(ui_->checkbox_download_album_covers->isChecked());
 
   s.endGroup();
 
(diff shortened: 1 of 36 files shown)


jonaski:
I wasn't aware that you can get cover art by album ID. Do you know if all all subsonic compatible server, supports this? In that case it doesn't need to be an option.

src/settings/subsonicsettingspage.cpp
```diff
@@ -90,6 +92,9 @@ void SubsonicSettingsPage::Load() {
       break;
   }
 
+  bool download_album_covers = ui_->checkbox_download_album_covers->isChecked();
```
jonaski, line 95: Juse use `ui_->checkbox_download_album_covers->isChecked()` directly instead of a variable here.

src/settings/subsonicsettingspage.cpp
```diff
@@ -109,17 +114,28 @@ void SubsonicSettingsPage::Save() {
...
+  s.setValue(kUseAlbumIdForAlbumCovers, ui_->checkbox_use_album_id_for_album_covers->isChecked());
   s.setValue(kServerSideScrobbling, ui_->checkbox_server_scrobbling->isChecked());
   if (ui_->auth_method_hex->isChecked()) {
     s.setValue(kAuthMethod, static_cast<int>(AuthMethod::Hex));
   }
   else {
     s.setValue(kAuthMethod, static_cast<int>(AuthMethod::MD5));
   }
+  
+  bool download_album_covers = ui_->checkbox_download_album_covers->isChecked();
```
jonaski, line 126: Ditto

src/settings/subsonicsettingspage.ui
```diff
@@ -171,6 +171,13 @@
         </property>
        </widget>
       </item>
+      <item>
+       <widget class="QCheckBox" name="checkbox_use_album_id_for_album_covers">
+        <property name="text">
+         <string>Use album id for album covers</string>
```
jonaski, line 177: ID with uppercase

src/subsonic/subsonicrequest.cpp
```diff
@@ -606,12 +606,19 @@ QString SubsonicRequest::ParseSong(Song &song, const QJsonObject &json_obj, cons
...
+      if (json_obj["coverArt"_L1].type() == QJsonValue::String) {
+        cover_id = json_obj["coverArt"_L1].toString();
+      }
+      else {
+        cover_id = QString::number(json_obj["coverArt"_L1].toInt());
+      }
     }
     else {
-      cover_id = QString::number(json_obj["coverArt"_L1].toInt());
+        cover_id = song_id;
```
jonaski, line 621: Wrong indent

src/subsonic/subsonicrequest.cpp
```diff
@@ -606,12 +606,19 @@ QString SubsonicRequest::ParseSong(Song &song, const QJsonObject &json_obj, cons
   if (json_obj.contains("genre"_L1)) genre = json_obj["genre"_L1].toString();
 
   QString cover_id;
-  if (json_obj.contains("coverArt"_L1)) {
-    if (json_obj["coverArt"_L1].type() == QJsonValue::String) {
-      cover_id = json_obj["coverArt"_L1].toString();
+  if (use_album_id_for_album_covers()) {
+    cover_id = album_id;
+  } else {
```
jonaski, line 611:
```suggestion
  }
  else {
```

src/translations/strawberry_ca_ES.ts
```diff
@@ -6684,6 +6684,10 @@ Esteu segur que voleu continuar?</translation>
       <source>Download album covers</source>
       <translation type="unfinished">Download album covers</translation>
     </message>
+    <message>
+      <source>Use album id for album covers</source>
```
jonaski, line 6688: ID

jonaski (changes requested):

jonaski:
The translations commit should be prefixed with "Translations: "

Leohige:
> I wasn't aware that you can get cover art by album ID. Do you know if all subsonic compatible server, supports this? In that case it doesn't need to be an option.

I've only used Navidrome, Subsonic, and Gonic. They all support retrieving the cover by the album ID. I don't know about the others, but they should support it if they follow the Subsonic API implementation.

src/settings/subsonicsettingspage.cpp
Leohige: Done

src/settings/subsonicsettingspage.cpp
Leohige: Done

src/settings/subsonicsettingspage.ui
Leohige: Done

src/subsonic/subsonicrequest.cpp
Leohige: Done

src/subsonic/subsonicrequest.cpp
Leohige: Done

src/translations/strawberry_ca_ES.ts
Leohige: Done

jonaski (approved):

PR #1698, opened 2025-03-12
Support arbitrarily large EBU R 128 loudness normalization

LebedevRI:
While i have fixed gstreamer's `volume` in
https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/5063 i did not see anything that followed after it was merged, namely, in https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/6222, the feature was moved, `"volume"` was reverted to only handle `x10` gain, and one needs to use `"volume-full-range"` instead to do arbitrary gain. So let's do that.

This, of course, requires run-time detection of the version of gstreamer base plugins that we are running with, specifically, we need version `1.24`.

### DIFF
--- a/src/engine/ebur128analysis.cpp
+++ b/src/engine/ebur128analysis.cpp
@@ -32,6 +32,7 @@
 #include <gst/gst.h>
 #include <gst/audio/audio-channels.h>
 #include <gst/app/gstappsink.h>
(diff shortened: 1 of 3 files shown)


LebedevRI:
@jonaski thank you!

Issue #351, opened 2020-01-16
Alsa backend device reference change every reboot


Ealrann:
First, thank you very much for strawberry. Finally a good audiophile music player on linux.

Well, I guess there is a linux solution to this problem, but I have to admit: I never really understood/trusted the alsa configuration files. So I tend to prefer to configure the front end software (strawberry).

**Is your feature request related to a problem? Please describe.**
When I reboot my computer, the device address of my (usb) backend change (hw:2,0 <-> hw:1,0). So I have to re-set it in the settings.

**Describe the solution you'd like**
Since I use my laptop at different locations, with different devices, the best solution for me would be: 
the capability to rank the devices to use, by name. Then, strawberry would use the most ranked device which is available.

**Describe alternatives you've considered**
Simply remember the name of the last used devices (rather than the location), and pick one randomly.

jonaski:
You need set index in modprobe for the card, there is some documentation here:
https://wiki.archlinux.org/index.php/Advanced_Linux_Sound_Architecture

PR #1711, opened 2025-03-28
AlsaDeviceFinder: Use card id instead of card index

corubba:
Like the card index, the card id is guaranteed to be unique. While the card index can easily change between reboots, the card id is based on the actual audio hardware and does not change between reboots; or even hardware changes. This makes using the card id preferable, because it will "just work" 99% of the time, and removes the need to force cards to have a specific index like in #351.

There is a corner case where card ids may change between reboots: If you have two (or more) of the same audio hardware in the system. But that should be rare enough, and requires explicit system configuration anyway, so using the "custom" option should work here.

If there is an previously-saved index-based ALSA device in the config, it will continue to work as-is and does not need to be migrated. There is only a small UI side-effect: Because the index-based device will no longer match any found id-based device, the settings window will show it as "custom". Simply selecting the ALSA device from the drop-down again will change it to the id-based device.

[This blog post](http://james.ahlstrom.name/alsa_names.html) has a nice summary of "index vs id", and while a bit aged it is still fairly accurate.

### DIFF
--- a/src/engine/alsadevicefinder.cpp
+++ b/src/engine/alsadevicefinder.cpp
@@ -98,9 +98,9 @@ EngineDeviceList AlsaDeviceFinder::ListDevices() {
       device.card = card;
       device.device = dev;
 
-      device.value = QStringLiteral("hw:%1,%2").arg(card).arg(dev);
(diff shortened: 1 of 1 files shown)


jonaski (approved):
Thanks

PR #1731, opened 2025-04-28
Use the coverArt field from song JSON if available

cquike:
This PR make sure that the coverArt JSON field in the subsonic song API takes precedence over the album or song ID.
As I understand it, if a song has a coverArt field then that one is the one to be used. As a fallback solution strawberry uses the song id. However if the setting "Use Album ID for album covers" is set, then the coverArt is ignored.

I have tried with the lms (https://github.com/epoupon/lms/tree/master) server and causes many errors since the album id is not mapped to the cover id (i.e., the only option for the getCoverArt method is to give the cover id).

My interpretation is that the "Use Album ID for album covers" uses album id as a fallback solution when the coverArt is not found in the song. If false, then the song id will be used as a fallback solution. That's the logic that this PR implements.

Thanks!

### DIFF
--- a/src/subsonic/subsonicrequest.cpp
+++ b/src/subsonic/subsonicrequest.cpp
@@ -376,6 +376,16 @@ void SubsonicRequest::AlbumSongsReplyReceived(QNetworkReply *reply, const QStrin
     created = QDateTime::fromString(object_album["created"_L1].toString(), Qt::ISODate).toSecsSinceEpoch();
   }
 
+  QString album_cover_id;
+  if (object_album.contains("coverArt"_L1)) {
+    if (object_album["coverArt"_L1].type() == QJsonValue::String) {
+      album_cover_id = object_album["coverArt"_L1].toString();
+    }
+    else {
+      album_cover_id = QString::number(object_album["coverArt"_L1].toInt());
+    }
+  }
+
   bool compilation = false;
   bool multidisc = false;
   SongList songs;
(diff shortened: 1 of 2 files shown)


jonaski:
As for as I understand, the problem was that using coverArt was requesting cover for each song, because each song had different coverArt, see #1636, so this will actually make it impossible to use album ID if coverArt exists which will re-introduce the issue again.
CC: @Leohige

cquike:
I see. It is a bit unfortunate that different Subsonic servers behave differently. So #1636  was introduced to make gonic and subsonic to behave sanely, while this PR would make lms to work.
OpenSubsonic (https://opensubsonic.netlify.app/docs/) is establishing itself as the new open standard, and gonic is a participant of it, and so is lms, so at least it would be nice if both behave in a compatible way. I cannot say for subsonic itself since it is not open source.
One possible solution is that the option  usealbumidforalbumcovers uses the album "coverArt" field to get the cover id, rather than calling getCoverArt with the album id. I will try to see whether that change would work. @Leohige : do you think that would work in your case?
Note that OpenSubsonic documentation clarifies the semantics of getCoverArt: https://github.com/opensubsonic/open-subsonic-api/discussions/140

cquike:
I have now modified the PR to ask for the coverArt field of the reponse to getAlbum if the  "Use album id for album covers" option is enabled. That should give the same benefit as #1636 while being compatible with the clarification of opensubsonic on the coverId for getCoverArt.
I have tested that this gives correct results with gonic, ampache and lms. @Leohige : if you could also test this patch would be great. Thanks!

cquike:
I have also tested now that it works with navidrome as well.

Leohige:
Unfortunately, I can't test this right now. I'm traveling, and my computer, which I use to compile the project, is at home.

src/subsonic/subsonicrequest.cpp
```diff
@@ -549,19 +559,14 @@ QString SubsonicRequest::ParseSong(Song &song, const QJsonObject &json_object, c
 
   QString cover_id;
   if (use_album_id_for_album_covers()) {
```
jonaski:
```suggestion
  if (use_album_id_for_album_covers() && !album_cover_id.isEmpty()) {
```

jonaski (changes requested):

jonaski:
@Leohige Can you test?

src/subsonic/subsonicrequest.cpp
cquike: I tested that and it works. I committed it to the branch.
Thanks for the review!

PR #1736, opened 2025-04-30
Feature/smartshuffle

federidos:
The original random shuffle in ShuffleMode::All and InsideAlbum has been replaced. Songs are now ordered based on:
-     Play count: tracks with fewer plays get higher priority.
-     Time since last played: older tracks are favored.
Exponential curves and dynamic adjustments are used to balance the ordering and reduce frequent repetition.

Previously, each time the app restarted, a completely new shuffle list was generated, ignoring recent playback history. This increased the chances of repeating recently played songs, while others could remain unplayed indefinitely. The new approach aims to mitigate this imbalance, especially in large playlists.

The ShuffleMode::Off and ShuffleMode::Albums remain unchanged.

### DIFF
--- a/src/playlist/playlist.cpp
+++ b/src/playlist/playlist.cpp
@@ -1,4 +1,4 @@
-/*
+ /*
  * Strawberry Music Player
  * This file was part of Clementine.
  * Copyright 2010, David Sansome <me@davidsansome.com>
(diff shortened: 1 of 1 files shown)


jonaski:
The unit tests fail. Also, I think this needs to be optional.
Formatting is wrong, indent should be 2 spaces, not 4. Variable names should be snake case, not camel case.

jonaski:
Closing this as there isn't any response.

PR #1737, opened 2025-05-01
Linux: enable startup notify

OlegAckbar:
It was very odd for me why Strawberry doesn't have any feedback when launching from application menu. Turns out its desktop file had "StartupNotify=false" for some reason?


jonaski:
I'm pretty sure this was done for a reason, but I don't remember why

OlegAckbar:
According to git blame this option was added 7 years ago... Maybe if there was a reason it's no longer relevant?

jonaski:
We can try

PR #1749, opened 2025-05-19
Allow transcoding to ALAC for iPods

mrpapersonic:
ALAC is the only suitable lossless codec for iPods; WAV and AIFF simply take up way too much space, usually around twice that of ALAC. rhythmbox has supported syncing ALAC to iPods for [a while now](https://mail.gnome.org/archives/commits-list/2016-October/msg08018.html), but \<opinion\>it is an abysmal music player\</opinion\>.

a quick test with my own iPod 5.5 gen shows this patch seems to work fine. It uses whatever gstreamer provides for `audio/x-alac` (most likely ffmpeg, I don't know of any others supporting encoding alac)

### DIFF
--- a/src/core/song.cpp
+++ b/src/core/song.cpp
@@ -1209,6 +1209,7 @@ QString Song::TextForFiletype(const FileType filetype) {
     case FileType::CDDA:        return u"CDDA"_s;
     case FileType::SPC:         return u"SNES SPC700"_s;
     case FileType::VGM:         return u"VGM"_s;
+    case FileType::ALAC:        return u"ALAC"_s;
     case FileType::Stream:      return u"Stream"_s;
     case FileType::Unknown:
     default:                    return QObject::tr("Unknown");
@@ -1241,6 +1242,7 @@ QString Song::ExtensionForFiletype(const FileType filetype) {
     case FileType::IT:          return u"it"_s;
     case FileType::SPC:         return u"spc"_s;
     case FileType::VGM:         return u"vgm"_s;
+    case FileType::ALAC:        return u"m4a"_s;
     case FileType::Unknown:
     default:                    return u"dat"_s;
   }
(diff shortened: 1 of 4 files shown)


src/core/song.cpp
```diff
@@ -1323,6 +1327,8 @@ Song::FileType Song::FiletypeByMimetype(const QString &mimetype) {
   if (mimetype.compare("audio/x-spc"_L1, Qt::CaseInsensitive) == 0) return FileType::SPC;
   if (mimetype.compare("audio/x-vgm"_L1, Qt::CaseInsensitive) == 0) return FileType::VGM;
 
+  // XXX: ALAC? Does gstreamer just give back audio/mpeg like it does for MP3 and AAC?  --paper
```
jonaski, line 1330: Think it should be `audio/x-alac`, but please test a file and check.

src/core/song.cpp
```diff
@@ -1380,6 +1387,10 @@ Song::FileType Song::FiletypeByExtension(const QString &ext) {
   if (ext.compare("spc"_L1, Qt::CaseInsensitive) == 0) return FileType::SPC;
   if (ext.compare("vgm"_L1, Qt::CaseInsensitive) == 0) return FileType::VGM;
 
+  // this can also be mp4/aac, etc, so commenting for now  --paper
```
jonaski, line 1390: If the extension is the same, than it's not much we can do here anyway, so just remove the comments.

jonaski (changes requested):
Please capitalize the commit messages, ie.: _Song: Add ALAC song type_

mrpapersonic:
What capitalization would you prefer for the gpoddevice commit? "GPodDevice" or something similar?

src/core/song.cpp
mrpapersonic: Yep, that's right.

jonaski (approved):

PR #1750, opened 2025-05-20
Discord Rich Presence: Display album cover art as the large image

mrpapersonic:
This is quite a bit more appealing than just the simple Strawberry logo :)

Example of what this looks like from within Discord:

image


jonaski:
I think we need to do this differently as this will request covers for each song change in addition to the original request if the cover was missing.
Have you checked if Discord accepts data URI scheme?

mrpapersonic:
I haven't, but I'm not sure if it will work, given that the image URL is limited to 128 characters (640k ought to be enough for anybody), I don't think it would fit.

Possibly we could also make it an option. I took that approach in wgsdk (my winamp plugin for discord presence) as well.

mrpapersonic:
> I think we need to do this differently as this will request covers for each song change in addition to the original request if the cover was missing. Have you checked if Discord accepts data URI scheme?

~~It looks like Discord *does* support [the `file://` URI scheme](https://github.com/TheQwertiest/foo_discord_rich/blob/44dbbe6178ce44ac1c8692db9883a19aab175eb8/foo_discord_rich/artwork/uploader.cpp#L71), so we could save the cover data to a temp file (png, jpg, webp, pick your poison) and let Discord load from that.~~

Nevermind, the code in question was to support uploading to an external website that Discord can then download the image from.

AzimovParviz:
> I haven't, but I'm not sure if it will work, given that the image URL is limited to 128 characters (640k ought to be enough for anybody), I don't think it would fit.
> 
> Possibly we could also make it an option. I took that approach in wgsdk (my winamp plugin for discord presence) as well.

I don't know about the RPC strawberry is using, but in Game SDK and Social SDK (replacement for deprecated Game SDK) you can only have URLs. You could potentially upload assets to Discord Application's Rich presence assets (there's endpoints for uploading and deleting) dynamically, but I personally couldn't get it to work (assets get uploaded but there's some time that needs to pass before it will show up which is annoying if you are listening to music, which could be as short as below 1 minute in length)

mrpapersonic:
Coming back to this, and from @AzimovParviz's input, I don't think there's a way to work around having to make a request(s) any time a song is played. Discord caches the files on the server side (i.e. `file://` links Won't Work™), which means we would need a public-facing HTTP server.

So making an option to disable it really seems like the only way to get rid of unnecessary requests.

> deprecated Game SDK

Why can't Discord just stick to a library already, FFS

AzimovParviz:
Well it's not exactly deprecated. It's legacy, e.g it won't receive any more updates and probably no more support if your game/app uses it , but the API will be there to use for the foreseeable future

rniii:
Does this fallback to the MusicBrainz cover art? If not, `https://coverartarchive.org/release/{release_mbid}/front` is a reasonable fallback for tracks with MBID metadata

https://musicbrainz.org/doc/Cover_Art_Archive/API#/release/%7Bmbid%7D/front

jonaski:
I suggest that we add a remote url to the songs table, when a new cover is looked up and stored, we also save the remote URL for that cover for Discord use. For Tidal, Spotify, etc we can store the remote URL also. As a fallback we can try MusicBrainz if there is a release_mbid for the album.

PR #1755, opened 2025-05-23
Images: Improved strawbs.png

pvdosev:
I edited the image so the strawberry leaves aren't cut off. It was slightly bothering me :)


pvdosev:
A screenshot of the new version of the image.
Here's a screenshot with how it looks like now.

Isnapple43:
It would be nice if this could be merged so I can rework `strawbs` to work with the KDE dark theme.

jonaski:
This removes the transparent background.

PR #1779, opened 2025-07-08
Support for Artist Sort Tags

x9foo:


data/schema/schema-21.sql
```diff
@@ -0,0 +1,57 @@
+DROP INDEX IF EXISTS idx_artistsort;
+
+DROP INDEX IF EXISTS idx_albumartistsort;
+
+ALTER TABLE songs ADD COLUMN artistsort TEXT;
```
jonaski, line 5: Please use `%allsongstables` to reduce the number of entries, and that will also update the device song tables.
See https://github.com/strawberrymusicplayer/strawberry/blob/a47531d4ce020b271f848fb821cb8a57e6f07488/src/core/database.cpp#L415

jonaski: You also need to add the columns to device-schema.sql https://github.com/strawberrymusicplayer/strawberry/blob/master/data/schema/device-schema.sql identical to songs, otherwise devices will fail.

src/constants/collectionsettings.h
```diff
@@ -50,6 +50,13 @@ constexpr char kOverwriteRating[] = "overwrite_rating";
 constexpr char kDeleteFiles[] = "delete_files";
 constexpr char kLastPath[] = "last_path";
 
+enum class SortShowArtists {
+  AsIs = 1,
+  SortSkipArticleShowArtist = 2,
```
jonaski, line 55:
```suggestion
 SkipArticles = 2,
```

src/constants/collectionsettings.h
```diff
@@ -50,6 +50,13 @@ constexpr char kOverwriteRating[] = "overwrite_rating";
 constexpr char kDeleteFiles[] = "delete_files";
 constexpr char kLastPath[] = "last_path";
 
+enum class SortShowArtists {
```
jonaski, line 53: I suggest different names so we can use the same enum for all sort tags:
`SortBehaviour`

src/constants/collectionsettings.h
```diff
@@ -50,6 +50,13 @@ constexpr char kOverwriteRating[] = "overwrite_rating";
...
+  AsIs = 1,
+  SortSkipArticleShowArtist = 2,
+  SortByTagShowArtist = 3,
```
jonaski, line 56: This name was a bit confusing to me, also don't limit the enum to artist.

```suggestion
  UseSortTagForSort = 3,
```

src/constants/collectionsettings.h
```diff
@@ -50,6 +50,13 @@ constexpr char kOverwriteRating[] = "overwrite_rating";
...
+  SortByTagShowTag = 4
```
jonaski, line 57:
```suggestion
  UseSortTagForDisplayAndSort = 4
```

x9foo:
Thanks for reviewing. I agree with all your comments. Handling the other sort tags as well certainly makes sense (I just hope that the ui doesn't get too messy).

What's the best way forward now? Since the scope changes quite a bit I propose the following:

- I create a new PR "Database Schema Version 21" that only includes my sort tag additions as well as your proposed bpm, mood, initial_key additions. **UPDATE**: [Change prepared](https://github.com/x9foo/strawberry/commit/8f6c0aa23a800162fa075bd32ae9b2507de33e5f)
- I create a new PR "Support for Sort Tags" where I basically replay and extend my changes from this PR and integrate your comments
- We leave this PR as reference for now and close it when the other two are done

That way we could get the db changes early into master and you can work on bpm, mood, initial_key without depending on me (I won't be able to work on the sort tags next week). What do you think?

jonaski:
You don't need to create a new PR, but you can add the schema update first if you want.

data/schema/schema-21.sql
x9foo: Resolved in https://github.com/strawberrymusicplayer/strawberry/pull/1781

src/constants/collectionsettings.h
x9foo: Resolved with https://github.com/strawberrymusicplayer/strawberry/pull/1788

src/constants/collectionsettings.h
x9foo: Resolved with https://github.com/strawberrymusicplayer/strawberry/pull/1788

src/constants/collectionsettings.h
x9foo: Resolved with https://github.com/strawberrymusicplayer/strawberry/pull/1788

src/constants/collectionsettings.h
x9foo: Resolved with https://github.com/strawberrymusicplayer/strawberry/pull/1788

PR #1785, opened 2025-07-29
Fixing behavior of error dialog

JohnMcLaren:
Ignore all error messages while the window is minimized.


jonaski:
> This fix finally prevents error messages on network errors for playing streams.

The changes doesn't do what the commit message says, it actually ignores any error messages while the window is minimized, not just network errors.

jonaski:
Ignoring all errors isn't a good idea since it makes it very difficult to diagnose a problem.

PR #1788, opened 2025-08-01
Support for Sort Tags

x9foo:
(Note: This PR replaces https://github.com/strawberrymusicplayer/strawberry/pull/1779 which offered support for "artist sort" and "album artist sort". This new PR now adds support for **all** common sort tags, as suggested by Jonas. PR https://github.com/strawberrymusicplayer/strawberry/pull/1781 has been split off to prepare the database schema for the new fields. The remaining text and screenshots have been updated to reflect the new features.)

# Overview

This PR adds support for sorting a collection by sort names often included as special tags in song files. This allows, for example, to see "The Beatles" sorted as "Beatles, The".

Strawberry already has some support for filtering articles from artist names before sorting, but this has limitations like no support for languages other than English. A more reliable way is to use provided data since the [rules for correct sort names can be quite complex](https://musicbrainz.org/doc/Style/Artist/Sort_Name). If you tag your files with MusicBrainz, for example, then sort information is stored in tags like "artist sort", "album artist sort" or "composer sort".

Technically, the PR adds support for reading, editing, and applying the following vorbis (id3v2) sort tags:

- ALBUMARTISTSORT (TSO2)
- ALBUMSORT (TSOA)
- ARTISTSORT (TSOP)
- COMPOSERSORT (TSOC)
- PERFORMERSORT (-)
- TITLESORT (TSOT)

See e.g. https://docs.mp3tag.de/mapping-table/ or https://picard-docs.musicbrainz.org/downloads/MusicBrainz_Picard_Tag_Map.html for tag definitions.

This PR also addresses requests in the following forum posts:

- https://forum.strawberrymusicplayer.org/topic/696/articles-in-artist-names-when-sorting-library
- https://forum.strawberrymusicplayer.org/topic/5757/sort-tags
- https://forum.strawberrymusicplayer.org/topic/5803/skip-leading-articles-when-sorting-albums
- https://forum.strawberrymusicplayer.org/topic/1265/sorting-by-tsop

# User Visible Changes

## Collection

The **Settings|Collection** page now has a combo box for choosing the sort behavior. Previous versions of strawberry had a checkbox "Skip leading articles ("the", "a", "an") when sorting artist names". This has been removed, but the underlying behavior is still available as one of the combobox options. The sort behavior applies to any name tag that may have a corresponding sort tag: album, artist, album artist, title, composer, and performer.

collectionsettings

There are four settings available:

combobox

Option 1 (**Sort and show name as is**) uses the name as is. It is equivalent to the unchecked state of the "Skip leading articles..." checkbox in previous versions of strawberry. Here is an example of the result when grouping by artist name:

option-1

Option 2 (**Skip articles "The, A, An" for sorting but show name as is**) is the default and the same as the checkbox "Skip leading articles..." in earlier versions. Note that this option does not correctly handle non-English names as can be seen by the entry for "Die Ärzte".

option-2

Options 1 and 2 provide full backwards compatibility. They can be used, for example, if your collection does not have sort tags, you do not want to use them, or you are just happy with the current behavior.

Option 3 and 4 provide the new functionality by using sort tags. Option 3 (**Use sort tag for sorting but show name as is**) uses the sort tag for sorting internally, but shows the original name.

option-3

Option 4 (**Use sort tag for sorting and display**) uses the sort tag internally and also for display.

option-4

Another example: if you have a collection of classical music and want to sort by composer last name you can do so by selecting **Group by|Advanced grouping** 

advanced-grouping

and "Composer" in the following dialog:

advanced-grouping-dialog

## Playlists

There are six new columns that can be enabled in the context menu of playlists:

playlistmenu

An example with two new columns enabled:

playlist

Some custom playlist settings like column positions or size adjustments are reset to defaults in this version. This is due to internal data structure changes.

## Smart Playlists

Smart playlists can use search terms with the new fields.

smartplsearch

The new fields can then be used for sorting the results as well.

smartplsort

## Edit Track Information

The **Edit Track Information** dialog allows to edit the new fields for supported file formats (e.g. flac and mp3). Changes are written back to the song files and the collection database is updated as well.

edittrack

## Context View

The custom text settings for the context view can now make use of the new fields. The popup menu in the **Settings|Context** page has additional entries for that:

contextsettings

Here is an example setting using the **artistsort** field.

context1

context2

## Migrating Existing Collections

When the new version of strawberry detects an older collection database on startup, it offers to rescan the collection.

rescan

This should be confirmed in order to read information for the new tags from song files.

# Implementation Details

The steps necessary to implement this were:

- Add new fields to database schema
- Increment database schema from 20 to 21
- Add migration script for database upgrade
- Add new fields to Song class
- Read/write vorbis and id3v2 tags in Tagreader and set/get corresponding Song fields
- Implement new sort options in CollectionModel
- Extend settings ui
- Extend playlist view
- Extend smart playlist ui
- Extend edit track dialog
- Trigger dialog offering collection rescan on startup
- Add unit tests

## View Playlist

I wanted to keep the new columns close to their related ones, i.e., "artist sort" should come next to "artist". It seems that the column positions are directly mapped to the "Column" enum in src/playlist/playlist.h. Appending to that enum as suggested by the comment on top of the enum has the effect that all sort tags by default appear at the very end of the popup menu and the right side of the playlist. I therefore merged the new fields into the enum and incremented the playlist state version from 1 to 2 to get sort columns next to their "original" column. That will discard old stored playlist state in config files.

# Limitations

- Tested only on Fedora 42
- Tested only with flac and mp3 files
- Unit tests are quite limited in scope, most tests where done manually

# Tests

I have run the following manual tests with a flac and mp3 collection of about 1,000 albums / 16,000 songs:

- Verify schema migration is triggered on start up, and adds / reads new fields
- Verify all four options for sorting the collection work
- Verify empty sort fields do not crash strawberry
- Verify playlist columns can be switched on, can be used to sort, and can be edited
- Verify edit track information dialog allows to edit new fields
- Verify smart playlist allows to use new fields for selection and sorting

In addition, I used the following two bash scripts to generate flac files with sort tags for testing.

gen-all-flac.sh

```bash
#!/usr/bin/env bash

set -euo pipefail

# 2592 songs
N_ALBUMS=2
N_ARTISTS=3
N_TRACKS=12

mkdir -p ./tmp

N=0
for letter in {Á,Ä,Å,Ã,Æ,Ġ,Ñ,Ş,Ÿ,Ž,{A..Z}}; do
    for ((artist = 1; artist <= N_ARTISTS; artist++)); do
        for ((album = 1; album <= N_ALBUMS; album++)); do
            for ((track = 1; track <= N_TRACKS; track++)); do
                echo $letter $artist $album $track
            done
        done
    done
done | xargs -n 4 -P $(nproc) ./gen-one-flac.sh
```

gen-one-flac.sh

```bash
#!/usr/bin/env bash

set -euo pipefail

letter=$1
artist=$2
album=$3
track=$4
# echo $letter $artist $album $track

full_artist="The ${letter} artist ${artist}"
full_artistsort="${letter} artist ${artist}, The"
full_album="The album ${album} by ${full_artist}"
full_albumsort="album ${album}, The by ${full_artist}"
full_title="The Title ${track} on ${full_album}"
full_titlesort="Title ${track}, The on ${full_album}"
full_composer="The ${letter} composer ${artist}"
full_composersort="${letter} composer ${artist}, The"
full_performer="The ${letter} performer ${artist} (piano)"
full_performersort="${letter} performer ${artist}, The (piano)"

exec ffmpeg -nostdin -loglevel error -y -f lavfi -t 1ms -i anullsrc=channel_layout=stereo:sample_rate=44100 \
    -metadata "ALBUMARTIST=${full_artist}" \
    -metadata "ALBUMARTISTSORT=${full_artistsort}" \
    -metadata "ALBUM=${full_album}" \
    -metadata "ALBUMSORT=${full_albumsort}" \
    -metadata "ARTIST=${full_artist}" \
    -metadata "ARTISTSORT=${full_artistsort}" \
    -metadata "COMPOSER=${full_composer}" \
    -metadata "COMPOSERSORT=${full_composersort}" \
    -metadata "PERFORMER=${full_performer}" \
    -metadata "PERFORMERSORT=${full_performersort}" \
    -metadata "DATE=2000" \
    -metadata "DISCNUMBER=1" \
    -metadata "DISCTOTAL=1" \
    -metadata "TITLE=${full_title}" \
    -metadata "TITLESORT=${full_titlesort}" \
    -metadata "TRACKNUMBER=$track" \
    ./tmp/$letter-$artist-$album-$track.flac
```

These scripts create songs that look like this:

testdata

### DIFF
--- a/src/playlist/playlistview.cpp
+++ b/src/playlist/playlistview.cpp
@@ -87,6 +87,7 @@ constexpr int kGlowIntensitySteps = 24;
 constexpr int kAutoscrollGraceTimeout = 30;  // seconds
 constexpr int kDropIndicatorWidth = 2;
 constexpr int kDropIndicatorGradientWidth = 5;
+constexpr int kHeaderStateVersion = 2;
 }  // namespace
 
 PlaylistView::PlaylistView(QWidget *parent)
@@ -131,7 +132,6 @@ PlaylistView::PlaylistView(QWidget *parent)
       cached_current_row_row_(-1),
       drop_indicator_row_(-1),
       drag_over_(false),
-      header_state_version_(1),
       column_alignment_(DefaultColumnAlignment()),
       rating_locked_(false),
       dynamic_controls_(new DynamicPlaylistControls(this)),
@@ -217,12 +217,18 @@ void PlaylistView::SetItemDelegates() {
   setItemDelegate(new PlaylistDelegateBase(this));
 
   setItemDelegateForColumn(static_cast<int>(Playlist::Column::Title), new TextItemDelegate(this));
+  setItemDelegateForColumn(static_cast<int>(Playlist::Column::TitleSort), new TagCompletionItemDelegate(this, collection_backend_, Playlist::Column::TitleSort));
   setItemDelegateForColumn(static_cast<int>(Playlist::Column::Album), new TagCompletionItemDelegate(this, collection_backend_, Playlist::Column::Album));
+  setItemDelegateForColumn(static_cast<int>(Playlist::Column::AlbumSort), new TagCompletionItemDelegate(this, collection_backend_, Playlist::Column::AlbumSort));
   setItemDelegateForColumn(static_cast<int>(Playlist::Column::Artist), new TagCompletionItemDelegate(this, collection_backend_, Playlist::Column::Artist));
+  setItemDelegateForColumn(static_cast<int>(Playlist::Column::ArtistSort), new TagCompletionItemDelegate(this, collection_backend_, Playlist::Column::ArtistSort));
   setItemDelegateForColumn(static_cast<int>(Playlist::Column::AlbumArtist), new TagCompletionItemDelegate(this, collection_backend_, Playlist::Column::AlbumArtist));
+  setItemDelegateForColumn(static_cast<int>(Playlist::Column::AlbumArtistSort), new TagCompletionItemDelegate(this, collection_backend_, Playlist::Column::AlbumArtistSort));
   setItemDelegateForColumn(static_cast<int>(Playlist::Column::Genre), new TagCompletionItemDelegate(this, collection_backend_, Playlist::Column::Genre));
   setItemDelegateForColumn(static_cast<int>(Playlist::Column::Composer), new TagCompletionItemDelegate(this, collection_backend_, Playlist::Column::Composer));
+  setItemDelegateForColumn(static_cast<int>(Playlist::Column::ComposerSort), new TagCompletionItemDelegate(this, collection_backend_, Playlist::Column::ComposerSort));
   setItemDelegateForColumn(static_cast<int>(Playlist::Column::Performer), new TagCompletionItemDelegate(this, collection_backend_, Playlist::Column::Performer));
+  setItemDelegateForColumn(static_cast<int>(Playlist::Column::PerformerSort), new TagCompletionItemDelegate(this, collection_backend_, Playlist::Column::PerformerSort));
   setItemDelegateForColumn(static_cast<int>(Playlist::Column::Grouping), new TagCompletionItemDelegate(this, collection_backend_, Playlist::Column::Grouping));
   setItemDelegateForColumn(static_cast<int>(Playlist::Column::Length), new LengthItemDelegate(this));
   setItemDelegateForColumn(static_cast<int>(Playlist::Column::Filesize), new SizeItemDelegate(this));
@@ -304,11 +310,24 @@ void PlaylistView::LoadHeaderState() {
 
   Settings s;
   s.beginGroup(PlaylistSettings::kSettingsGroup);
+  // since we use serialized internal data structures, we cannot read anything but the current version
+  const int header_state_version_ = s.value(PlaylistSettings::kStateVersion, 0).toInt();
   if (s.contains(PlaylistSettings::kState)) {
-    header_state_version_ = s.value(PlaylistSettings::kStateVersion, 0).toInt();
-    header_state_ = s.value(PlaylistSettings::kState).toByteArray();
+    if (header_state_version_ == kHeaderStateVersion) {
+      header_state_ = s.value(PlaylistSettings::kState).toByteArray();
+    } else {
+      // force header state reset since column indices may have changed between versions
+      header_state_.clear();
+    }
   }
-  if (s.contains(PlaylistSettings::kColumnAlignments)) column_alignment_ = s.value(PlaylistSettings::kColumnAlignments).value<ColumnAlignmentMap>();
+  if (s.contains(PlaylistSettings::kColumnAlignments)) {
+    if (header_state_version_ == kHeaderStateVersion) {
+      column_alignment_ = s.value(PlaylistSettings::kColumnAlignments).value<ColumnAlignmentMap>();
+    } else {
+      // force column alignment reset since column indices may have changed between versions
+      column_alignment_.clear();
+    }
+  }
   s.endGroup();
 
   if (column_alignment_.isEmpty()) {
@@ -329,7 +348,6 @@ void PlaylistView::SetHeaderState() {
 void PlaylistView::ResetHeaderState() {
 
   set_initial_header_layout_ = true;
-  header_state_version_ = 1;
   header_state_ = header_->ResetState();
   RestoreHeaderState();
 
@@ -347,9 +365,15 @@ void PlaylistView::RestoreHeaderState() {
 
     header_->SetStretchEnabled(true);
 
+    header_->HideSection(static_cast<int>(Playlist::Column::TitleSort));
+    header_->HideSection(static_cast<int>(Playlist::Column::ArtistSort));
+    header_->HideSection(static_cast<int>(Playlist::Column::AlbumSort));
     header_->HideSection(static_cast<int>(Playlist::Column::AlbumArtist));
+    header_->HideSection(static_cast<int>(Playlist::Column::AlbumArtistSort));
     header_->HideSection(static_cast<int>(Playlist::Column::Performer));
+    header_->HideSection(static_cast<int>(Playlist::Column::PerformerSort));
     header_->HideSection(static_cast<int>(Playlist::Column::Composer));
+    header_->HideSection(static_cast<int>(Playlist::Column::ComposerSort));
     header_->HideSection(static_cast<int>(Playlist::Column::Year));
     header_->HideSection(static_cast<int>(Playlist::Column::OriginalYear));
     header_->HideSection(static_cast<int>(Playlist::Column::Disc));
@@ -398,11 +422,6 @@ void PlaylistView::RestoreHeaderState() {
 
     set_initial_header_layout_ = false;
 
-  }
-
-  if (header_state_version_ < 1) {
-    header_->HideSection(static_cast<int>(Playlist::Column::Rating));
-    header_state_version_ = 1;
   }
 
   // Make sure at least one column is visible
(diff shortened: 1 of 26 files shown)


src/playlist/playlistview.cpp
```diff
@@ -304,11 +310,24 @@ void PlaylistView::LoadHeaderState() {
...
   Settings s;
   s.beginGroup(PlaylistSettings::kSettingsGroup);
+  // since we use serialized internal data structures, we cannot read anything but the current version
+  const int header_state_version_ = s.value(PlaylistSettings::kStateVersion, 0).toInt();
   if (s.contains(PlaylistSettings::kState)) {
-    header_state_version_ = s.value(PlaylistSettings::kStateVersion, 0).toInt();
-    header_state_ = s.value(PlaylistSettings::kState).toByteArray();
+    if (header_state_version_ == kHeaderStateVersion) {
+      header_state_ = s.value(PlaylistSettings::kState).toByteArray();
+    } else {
```
jonaski, line 318:
```suggestion
    }
    else {
```

src/playlist/playlistview.cpp
```diff
@@ -304,11 +310,24 @@ void PlaylistView::LoadHeaderState() {
 
   Settings s;
   s.beginGroup(PlaylistSettings::kSettingsGroup);
+  // since we use serialized internal data structures, we cannot read anything but the current version
+  const int header_state_version_ = s.value(PlaylistSettings::kStateVersion, 0).toInt();
```
jonaski, line 314:
```suggestion
  const int header_state_version = s.value(PlaylistSettings::kStateVersion, 0).toInt();
```

src/playlist/playlistview.cpp
```diff
@@ -304,11 +310,24 @@ void PlaylistView::LoadHeaderState() {
...
+      header_state_ = s.value(PlaylistSettings::kState).toByteArray();
+    } else {
+      // force header state reset since column indices may have changed between versions
+      header_state_.clear();
+    }
+  }
+  if (s.contains(PlaylistSettings::kColumnAlignments)) {
+    if (header_state_version_ == kHeaderStateVersion) {
+      column_alignment_ = s.value(PlaylistSettings::kColumnAlignments).value<ColumnAlignmentMap>();
+    } else {
```
jonaski, line 326:
```suggestion
    }
    else {
```

src/playlist/playlistview.cpp
```diff
@@ -304,11 +310,24 @@ void PlaylistView::LoadHeaderState() {
...
+      // force column alignment reset since column indices may have changed between versions
```
jonaski, line 327:
```suggestion
      // Force column alignment reset since column indices may have changed between versions
```

src/playlist/playlistview.cpp
```diff
@@ -304,11 +310,24 @@ void PlaylistView::LoadHeaderState() {
...
   s.beginGroup(PlaylistSettings::kSettingsGroup);
+  // since we use serialized internal data structures, we cannot read anything but the current version
+  const int header_state_version_ = s.value(PlaylistSettings::kStateVersion, 0).toInt();
   if (s.contains(PlaylistSettings::kState)) {
-    header_state_version_ = s.value(PlaylistSettings::kStateVersion, 0).toInt();
-    header_state_ = s.value(PlaylistSettings::kState).toByteArray();
+    if (header_state_version_ == kHeaderStateVersion) {
+      header_state_ = s.value(PlaylistSettings::kState).toByteArray();
+    } else {
+      // force header state reset since column indices may have changed between versions
```
jonaski, line 319:
```suggestion
      // Force header state reset since column indices may have changed between versions
```

jonaski (changes requested):
Isn't it possible to avoid resetting the column state if we but the new columns last in Playlist::Column?
I had to reset the columns not to long ago because I rewrote parts of StretchHeaderView and some users complained, so I think we should avoid it if possible.
So instead resetting, place the new columns last and hide the new columns in a `if (header_state_version_ < 2)`, or is it something preventing this now?

Also, remember to sort the tabs in collection settings and edit tag dialog if you add new widgets.

jonaski:
I've manually fixed the formatting issues and merged this since I want to get this part of the next release, I kept the resetting of the columns. Thanks for your contribution.

x9foo:
Thanks, Jonas. Sorry for being late. I'm fine with the current approach. I was looking into a way to generally solve the "column reset" problem. I have experimented with an extra mapping between the column enum and the logical index replacing all the static_casts. You can see the code at https://github.com/x9foo/strawberry/commit/acfc694915c5fa201a614ae0f2506487c2ea0889 . It is probably not worth to go in this direction since the risk of breaking things seems high to me. However, it exposed a minor issue that also exists in current master. Using the bimap with internal range checking triggers a crash when I run strawberry with an empty config, right-click on the playlist header and select "Reset columns to default":

```
10:39:23.762 DEBUG Playlist:1488                    sort::column_number  -1
terminate called after throwing an instance of 'boost::wrapexcept<std::out_of_range>'
  what():  bimap<>: invalid key
ANOM_ABEND auid=1000 uid=1000 gid=1000 ses=3 subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 pid=21773 comm="strawberry" exe="/home/mr/work/strawberry/build/Desktop-Debug/strawberry" sig=6 res=1
10:39:25: The process crashed.
```

See playlist.cpp:1488 at https://github.com/x9foo/strawberry/commit/acfc694915c5fa201a614ae0f2506487c2ea0889#diff-324146d4cb3bf1ebfeae8483ef3790322a68a244fd46b5c17c93ba93a837e769R1488

Although master doesn't crash for me, the out-of-range static cast is still unspecified behavior.

> Also, remember to sort the tabs in collection settings and edit tag dialog if you add new widgets.

I have checked both dialogs again, they look ok to me. Is there anything specific still missing?

x9foo:
The feature is broken now after commit https://github.com/strawberrymusicplayer/strawberry/commit/c4646531b0be279f79887c1990db6bd3da9aee5f on master. Although the commit message says "Refactor use of sort tags" the change definitely is more than a simple refactoring. You have removed the combobox setting and the underlying SortBehavior enum. That means it's not possible anymore to display the sort tag content. Also, the two checkboxes you introduced don't seem to change anything?! With my test data I always get the same collection view no matter which of the four possible checkbox combinations I select:

473463509-e31dacb3-3756-4e9c-aa22-b6d1dee331b4

jonaski:
I was going to reply to you, I had to remove the option as I found several issues with the new option:
- Articles were always removed from albums too
- When sorting using sort tags, and one or more songs is missing sort tags, sorting is broken because it doesn't fall back to normal tags.

I figured that stripping articles need to be a separate options independent of showing sort tags or normal tags.
If there is a use-case for showing sort tags, we can add that back as an inpendent option, I actually worked for hours on it, but I was not happy with the result, so I removed it for now. Right now I'm trying to get things in a state were I can release a new version.

x9foo:
It is for sure not easy to get this right, especially with inconsistently or incompletely tagged collections. I actually expected the new options to need some time to stabilize. That's why I liked your proposal with the SortBehavior enum. The first two cases where implemented with full backwards compatibility in mind so that we would not risk breaking existing functionality. 

In case of missing sort tags, I would have assumed that some simple fallbacks like  the approach with Song::effective_artistsort() and friends could work as these kinds of fallbacks were already used in other places as well. 

> If there is a use-case for showing sort tags, we can add that back as an inpendent option

Actually, that is my main use case. :-)

Anyway, do the new checkboxes currently work for you as expected or is it just on my side?

jonaski:
> It is for sure not easy to get this right, especially with inconsistently or incompletely tagged collections. I actually expected the new options to need some time to stabilize. That's why I liked your proposal with the SortBehavior enum. The first two cases where implemented with full backwards compatibility in mind so that we would not risk breaking existing functionality.
> 
> In case of missing sort tags, I would have assumed that some simple fallbacks like the approach with Song::effective_artistsort() and friends could work as these kinds of fallbacks were already used in other places as well.
> 
> > If there is a use-case for showing sort tags, we can add that back as an inpendent option
> 
> Actually, that is my main use case. :-)
> 
> Anyway, do the new checkboxes currently work for you as expected or is it just on my side?

It works as I intended, to use sort tags for sort when available, fallback to normal tags. Strip articles for sorting, not for display.

x9foo:
Ok, that makes sense to me now. Since I have sort tags in my files, the checkboxes don't make any difference. My expectation was different, because that behavior is not obvious to me in the ui (perhaps it helps to put a title on top of the two checkboxes like "When there are no sort tags..."?)

Issue #1794, opened 2025-08-08
Windows build failure due to missing getopt-win32 package


camilajenny:
- [x] I have checked the [FAQ](https://wiki.strawberrymusicplayer.org/wiki/FAQ) for answers.
- [x] I have checked the [Changelog](https://github.com/strawberrymusicplayer/strawberry/blob/master/Changelog) that the issue is not already fixed. 
- [x] I believe this issue is a bug, and not a general technical issue, question or feature requests that can be discussed on the [forum](https://forum.strawberrymusicplayer.org/). 

**Describe the bug**
The current `CMakeLists.txt` uses `find_package(getopt-win REQUIRED)`, but the `getopt-win32` package installed via vcpkg provides targets under the name `unofficial-getopt-win32`. This causes build failures or missing package errors when using vcpkg.

**To Reproduce**
1. Install `getopt-win32` package using vcpkg (`vcpkg install getopt-win32`).
2. Try building the project on Windows with vcpkg integration enabled.
3. CMake fails with: `Could not find a package configuration file provided by "getopt-win"`.

**Expected behavior**
CMake should find and link the installed `getopt-win32` package correctly using the appropriate package name and target.

**Screenshots:**
N/A

**System Information:**
 - Operating system: Windows 11
 - Strawberry Version: 1.2.11

**Additional context**
Renaming the CMake call to `find_package(unofficial-getopt-win32 REQUIRED)` and linking with `unofficial::getopt-win32::getopt` resolves the issue when using vcpkg. This aligns the project with vcpkg naming conventions and improves Windows compatibility.

See also the contributor's comment: https://github.com/strawberrymusicplayer/strawberry/issues/1715#issuecomment-2781029357 
@sboukortt

jonaski:
The name I use is the one proposed in the CMake support here:
https://github.com/lukeprince20/getopt-win/blob/1c5f0b829e6d116ac40f3dbc248a218acac41dab/getopt-win-config.cmake.in#L3
Never supported building Strawberry with vcpkg

PR #1795, opened 2025-08-08
update the getopt package name for windows

camilajenny:
Solves https://github.com/strawberrymusicplayer/strawberry/issues/1794


jonaski:
This will obviously break the Windows builds, so if "unofficial-getopt-win32" is really needed, then it needs to check for both names.

jonaski:
6513b3032b366363ad93ce7ef0da2c552213335f

PR #1805, opened 2025-08-24
Fix album shuffle playing previous queue entries after restart

### LINKED ISSUE #1623: Shuffle Albums restarts current album after stop
- [X] I have checked the [FAQ](https://wiki.strawberrymusicplayer.org/wiki/FAQ) for answers.
- [X] I have checked the [Changelog](https://github.com/strawberrymusicplayer/strawberry/blob/master/Changelog) that the issue is not already fixed. 
- [X] I believe this issue is a bug, and not a general technical issue, question or feature requests that can be discussed on the [forum](https://forum.strawberrymusicplayer.org/). 

**Describe the bug**
When in "Shuffle Albums" mode and stopping an album playing partly through, using pause or stop, and then restarting play later will cause the current track to complete but then Strawberry restarts the album from the beginning instead of moving on to the next song. This same behavior can be seen when closing and relaunching Strawberry.

In other words, Strawberry seems to be losing track of its place when in "Shuffle Albums" mode

**To Reproduce**
1. Load a playlist with at least one album containing multiple tracks
2. Set the "Shuffle Albums" mode on
3. Play the 2nd song in an album. Skip ahead to near the end of the song and hit pause
4. Close Strawberry
5. Open Strawberry and continue playing the song
6. When the song completes, Strawberry moves to the 1st song in the album instead of the 3rd

**Expected behavior**
1. Load a playlist with at least one album containing multiple tracks
2. Set the "Shuffle Albums" mode on
3. Play the 2nd song in an album. Skip ahead to near the end of the song
4. When the song completes, Strawberry moves to the 3rd song in the album as it would in "Shuffle Tracks" mode

**System Information:**
 - Operating system: MacOS 15.2
 - Strawberry Version: 1.2.3

**Additional context**
Closing/Reopening Strawberry is not necessary to trigger this behavior. I've seen it occur when Strawberry is open put left in pause mode on a song for a period of time. I've not kept track to see how long but it seems to be more than just a few minutes. It does not seem to specifically be triggered by the screen saver nor sleep mode.

AzimovParviz:
Fixes #1623

### DIFF
--- a/src/playlist/playlist.cpp
+++ b/src/playlist/playlist.cpp
@@ -2078,6 +2078,10 @@ void Playlist::ReshuffleIndices() {
   // Update current virtual index
   if (current_item_index_.isValid()) {
     current_virtual_index_ = static_cast<int>(virtual_items_.indexOf(current_item_index_.row()));
+    // If currently playing album song is not the first one, remove the first song to prevent https://github.com/strawberrymusicplayer/strawberry/issues/1623 from happening
+    if (current_virtual_index_ > 0) {
+      virtual_items_.removeAt(virtual_items_.indexOf(current_item_index_.row()) - 1);
+    }
   }
   else {
     current_virtual_index_ = -1;
--- a/src/playlist/playlist.cpp
+++ b/src/playlist/playlist.cpp
@@ -2079,7 +2079,7 @@ void Playlist::ReshuffleIndices() {
   if (current_item_index_.isValid()) {
     current_virtual_index_ = static_cast<int>(virtual_items_.indexOf(current_item_index_.row()));
     // If currently playing album song is not the first one, remove the first song to prevent https://github.com/strawberrymusicplayer/strawberry/issues/1623 from happening
-    if (current_virtual_index_ > 0) {
+    if (current_virtual_index_ > 0 && shuffle_mode == PlaylistSequence::ShuffleMode::Albums) {
       virtual_items_.removeAt(virtual_items_.indexOf(current_item_index_.row()) - 1);
     }
   }


AzimovParviz:
I've read that you don't have macOS maintainers so I thought I'll try closing one of the macOS issues (although this one seems to be platform-agnostic)

jonaski:
Thanks for looking into this, but the fix isn't correct, all songs in the playlist should to be in `virtual_items_`, the problem is probably in the reshuffle, it doesn't consider that the previous songs have been played.

AzimovParviz:
> Thanks for looking into this, but the fix isn't correct, all songs in the playlist should to be in `virtual_items_`, the problem is probably in the reshuffle, it doesn't consider that the previous songs have been played.

Yeah I noticed too that it shuffles often even before you get to hit play on the song. I am thinking it's somewhere on line 751 in playlist.cpp. I couldn't deduce further

AzimovParviz:
@jonaski so would something along the lines of this be more suitable? in the current state it's not functional yet but hopefully the direction I wanna take is clear. I was thinking that alternatively a more simple solution could be just moving the virtual index based on the current row's track number in its album in the Playlist::set_current_row(). Let me know if that makes more sense

AzimovParviz:
I mostly work on this in the evening past 9 pm after a whole work day. I am going to tell myself that it's the reason why I came up with that overcomplicated mess of a commit [9130760](https://github.com/strawberrymusicplayer/strawberry/pull/1805/commits/9130760a6145d1a2b7f57283b12d9a767c8a5099)

this one should be much more elegant and less intrusive

AzimovParviz:
@jonaski I just noticed that I forgot to make a commit with proper commit message format. Do you want me to do it (assuming the current implementation is good enough to be merged?)

jonaski:
Requesting the entire playlist from SQL on each reshuffle can block the GUI and should be unnecessary.
Also, using the skip feature seem like a workaround and not a proper fix.

jonaski:
Try with 000ba997fbb9625b198ad61035c7962ec7571023

PR #1813, opened 2025-09-10
discord: fixed timestamp update when seeking

7xnl:
When seeking in a song, `RichPresence::Seeked()` receives the new position in microseconds and is supposed to update the RPC timestamp with the new position in seconds. However, it actually converts the value to milliseconds, meaning that if, for example, you seek to 0:05 in a song, Discord will think you seeked to 83:20, or 5000 seconds from the beginning of the song.

This commit fixes this by simply dividing the microseconds value by one million instead of one thousand.


7xnl:
Just to demonstrate - this happens when I start playing the song and seek just a few seconds ahead.

PR #1854, opened 2025-11-10
Fix length filter not converting to nanoseconds

uninstall-your-browser:
I noticed that if I typed `length:<60` the filter would always say there were no results, until today when I kept adding 0s and noticed that `length:<60000000000` did what I wanted it to

### DIFF
--- a/src/filterparser/filterparser.cpp
+++ b/src/filterparser/filterparser.cpp
@@ -22,6 +22,7 @@
 
 #include <QString>
 
+#include "constants/timeconstants.h"
 #include "filterparser.h"
 #include "filtertreenop.h"
 #include "filtertreeand.h"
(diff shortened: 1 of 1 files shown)


src/filterparser/filterparser.cpp
```diff
@@ -322,7 +322,8 @@ FilterTree *FilterParser::createSearchTermTreeNode(const QString &column, const
     else if (Song::kInt64SearchColumns.contains(column, Qt::CaseInsensitive)) {
       qint64 number = 0;
       if (column == "length"_L1) {
-        number = ParseTime(value);
+        // Convert seconds to nanoseconds
+        number = ParseTime(value) * 1'000'000'000;
```
jonaski: Include constants/timeconstants.h and use kNsecPerSec, then you can also drop the comment

uninstall-your-browser: fixed

jonaski (approved):

PR #1913, opened 2025-12-20
Add genre metadata support for streaming services

RobStanfield:
## Summary

- Add genre extraction for Qobuz and Spotify tracks during catalog refresh
- Add "Fetch metadata from service" context menu for existing playlist items
- Enable inline tag editing (genre, etc.) for streaming service tracks
- Enable multi-track "Set value to" for streaming service tracks
- Add request throttling for metadata fetching to avoid rate limiting

## Details

This PR adds comprehensive genre metadata support for streaming services (Qobuz, Spotify, Tidal, Subsonic):

**Genre extraction during catalog refresh:**
- Qobuz: Extracts genre from `album.genre.name` in the API response
- Spotify: Extracts genre from artist's genres array (requires additional API call)

**Fetch metadata for existing tracks:**
- New right-click menu option "Fetch metadata from service" 
- Works on selected Qobuz/Spotify tracks that don't have genre
- Throttled queue (200ms between requests) to avoid rate limiting

**Inline editing support:**
- `Song::IsEditable()` now returns true for streaming services
- `Playlist::setData()` handles streaming tracks (saves to playlist DB)
- `MainWindow::SelectionSetValue()` handles streaming tracks for multi-select
- `EditTagDialog` properly saves metadata changes for streaming tracks

## Test plan

- [ ] Add Qobuz/Spotify favorites and verify genre column is populated
- [ ] Right-click existing tracks without genre → "Fetch metadata from service"
- [ ] Verify genre appears after fetching
- [ ] Right-click genre column → "Edit tag genre" → change value → verify it persists
- [ ] Select multiple tracks → right-click → "Set genre to 'X'" → verify all updated
- [ ] Restart Strawberry and verify edited genres persist

### DIFF
--- a/src/core/mainwindow.cpp
+++ b/src/core/mainwindow.cpp
@@ -173,9 +173,12 @@
 #endif
 #ifdef HAVE_SPOTIFY
 #  include "spotify/spotifyservice.h"
+#  include "spotify/spotifymetadatarequest.h"
 #  include "constants/spotifysettings.h"
 #endif
 #ifdef HAVE_QOBUZ
+#  include "qobuz/qobuzservice.h"
+#  include "qobuz/qobuzmetadatarequest.h"
 #  include "constants/qobuzsettings.h"
 #endif
 
@@ -379,8 +382,10 @@ MainWindow::MainWindow(Application *app,
       playlist_add_to_another_(nullptr),
       playlistitem_actions_separator_(nullptr),
       playlist_rescan_songs_(nullptr),
+      playlist_fetch_metadata_(nullptr),
       track_position_timer_(new QTimer(this)),
       track_slider_timer_(new QTimer(this)),
+      metadata_queue_timer_(new QTimer(this)),
       keep_running_(false),
       playing_widget_(true),
 #ifdef HAVE_DBUS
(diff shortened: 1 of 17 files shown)


src/core/mainwindow.cpp
```diff
@@ -2214,6 +2234,142 @@ void MainWindow::RescanSongs() {
...
+    if (!item) continue;
+
+    const Song &song = item->EffectiveMetadata();
+    const Song::Source source = song.source();
+    const QPersistentModelIndex persistent_index = QPersistentModelIndex(source_index);
+    QString track_id;
+
+#ifdef HAVE_QOBUZ
+    if (source == Song::Source::Qobuz) {
+      track_id = song.url().path();
```
jonaski: Use `song.song_id()`

src/core/mainwindow.cpp
```diff
@@ -2214,6 +2234,142 @@ void MainWindow::RescanSongs() {
...
+        qLog(Error) << "Failed to fetch Qobuz metadata: No track ID in URL";
+        continue;
+      }
+      metadata_queue_.append({source, track_id, persistent_index});
+    }
+#endif
+
+#ifdef HAVE_SPOTIFY
+    if (source == Song::Source::Spotify) {
+      QString url_str = song.url().toString();
```
jonaski: Use `song.song_id()`

jonaski (changes requested):
I don't mind adding support for editing metadata, but I think the "Fetch metadata from service" should update the entire metadata, not just genre

src/core/mainwindow.cpp
```diff
@@ -2214,6 +2234,142 @@ void MainWindow::RescanSongs() {
...
+      if (url_str.startsWith(u"spotify:track:"_s)) {
+        track_id = url_str.mid(14);  // Length of "spotify:track:"
```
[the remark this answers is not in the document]
jonaski: `song.song_id()` should be used instead

RobStanfield:
Addressed the feedback:

1. **Fetch all metadata, not just genre** - `MetadataRequest` classes now return full `Song` objects with title, artist, album, albumartist, genre, track, disc, year, duration, and cover art. The callback in `MainWindow` applies all non-empty fields to the playlist item.

2. **Use `song_id()`** - Implemented with URL fallback since `song_id()` is empty when loaded from playlist (not persisted). Code tries `song_id()` first, falls back to URL parsing if empty.

Tested both Qobuz and Spotify - metadata fetch working correctly for multiple tracks.

jonaski (approved):

xstable:
May I ask a question here - sorry, if it's may be off-topic.. but it's regarding to genre fetching. 
If I currently fetch Metadata, I never get any genre. Should I get the genre, or is there any way to enable this?
I've currently spotify (free version) connected.

PR #1922, opened 2025-12-26
Add tree view mode to the Files tab

alexopus:
Add a tree view mode to the file browser, providing an alternative way to navigate files alongside the existing list view.

- Dual-mode file browser: toggle between list view and tree view
   - Toggle button switches between views with view-specific UI
   - List view: same as before (original view) <-- default
   - Tree view: add/remove root buttons

- Tree view implementation:
  - Virtual root structure with lazy-loaded filesystem
  - Uses `SimpleTreeModel<FileViewTreeItem>` with item types: `Root`, `VirtualRoot`, `Directory`, `File`
  - Each configured root path appears as a separate top-level item
  - Lazy loading of directory contents via `canFetchMore()`/`fetchMore()`

- Settings persistence: Configured roots and active view mode are stored in settings

Rather than adding the entire `/` as root, I opted for virtual roots, that allows adding multiple specific locations, to reduce parsing time and clutter.

Screenshots:

Button to switch to the tree view added in top right of the original Files tab:

image

The tree view offers add/remove button for virtual roots:

image

Adding works via standard open dialog. Upon addition of the virtual root, it works as standard tree view:

image

The code was mainly written by Claude Code. The C++ part looks reasonable to me. But I am not familiar with Qt, so perhaps the choice of the components is not optimal.

I've been using this feature locally for a couple of weeks now and it works well. Would be nice to have it upstream.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -823,6 +823,8 @@ set(SOURCES
 
   src/fileview/fileview.cpp
   src/fileview/fileviewlist.cpp
+  src/fileview/fileviewtree.cpp
+  src/fileview/fileviewtreemodel.cpp
 
   src/device/devicemanager.cpp
   src/device/devicelister.cpp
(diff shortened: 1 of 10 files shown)


alexopus:
I removed some unnecessary duplication and added drag support for directories out of tree view (was only done for files before)

PR #1941, opened 2026-01-03
Include .webp in allowed image extensions

marcusmueller:
Modern Qt can read and write webp out of the box, no use excluding that.

(so does my cover scanning pipeline, so there's at least one user of this :) )

### DIFF
--- a/src/constants/filefilterconstants.h
+++ b/src/constants/filefilterconstants.h
@@ -33,7 +33,7 @@ constexpr char kFileFilter[] =
     "*.mod *.s3m *.xm *.it "
     "*.spc *.vgm";
 
-constexpr char kLoadImageFileFilter[] = QT_TRANSLATE_NOOP("FileFilter", "Images (*.png *.jpg *.jpeg *.bmp *.gif *.xpm *.pbm *.pgm *.ppm *.xbm)");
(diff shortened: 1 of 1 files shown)


jonaski (approved):

BSDKaffee:
WEBP support is not exactly an OOB experience. Even Qt5 supports webp, but for both Qt5 and Qt6, qtimageformats must be installed with webp support enabled to handle this format. qtimageformats is a plugin, so it is only needed at runtime.

marcusmueller:
@BSDKaffee I wasn't aware of that. Hard to say what a sensible way forward here is? I'd really like to keep the ability to select WebP images. At the same same, offering to select a file on a machine that doesn't support it seems bad. 

Options:
- Do nothing (makes people who select webp but don't haven't got qtimageformats sad; but minimizes maintenance effort)
- Revert this commit (makes me and other webp users sad)
- Declare qtimageformats a run-time dependency (makes maintainer do work; might make downstream packagers unhappy, if they don't work off project-supplied packaging templates, they might just miss that)
- Check for presence of WebP support at runtime and adjust file name filters accordingly (makes whoever has to write that slightly sad)

thoughts?

marcusmueller:
(note that instead of hard-coding file extensions, we could use `QFileDialgo::setMimeTypeFilters(…)` with the list of MIME types returned from `QImage{Reader,Writer}::supportedImageFormats()`, then read the thus-computed NameFilters, then combine them into "any image format" name filter, prepend that, and append a "all files" filter.)

PR #1952, opened 2026-01-05
Fix "Show in file browser" opens the file instead of directory

peterhoeg:
(KDE), clicking "Show in file browser" makes dolphin open the *file*
rather than the directory in which it resides.
    
This has the unfortunate side effect and then starting playback of the
file in question in the default media player.
    
This patch just opens the directory rather than the file (as it should).

On a completely separate note - is there any particular reason that
`OpenInFileManager` tries to be so clever rather than just invoking
`xdg-open` and let that figure out what to do?

### DIFF
--- a/src/utilities/filemanagerutils.cpp
+++ b/src/utilities/filemanagerutils.cpp
@@ -85,7 +85,7 @@ void OpenInFileManager(const QString &path, const QUrl &url) {
            command.startsWith("dolphin"_L1) ||
            command.startsWith("konqueror"_L1) ||
            command.startsWith("kfmclient"_L1)) {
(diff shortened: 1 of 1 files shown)


jonaski:
The point of the `dolphin --select` command is to open the folder and select the file, this works fine when testing here.
This change will break this feature. I think you need to debug this on your system, run `dolphin --select <file>` and check if that works outside of strawberry first, if that doesn't work either, then the bug isn't in strawberry.

peterhoeg:
It’s turns out it’s a dolphin bug.

If dolphin is told to “Begin in split view mode”, --select is broken as it opens the file instead of selecting it.

And because it’s a dual view, it gets opened twice…

I’ll see how I can fix dolphin. Thank you for strawberry!

PR #1953, opened 2026-01-05
feat: create compile_commands.json for clangd

peterhoeg:
cmake has the option to automatically create `compile_commands.json` for
use by clangd in order to enable code navigation, context sensitive help
and so on which is really nice when trying to navigate an unfamiliar
codebase.
    
This feature depends on cmake 3.5 but as we already require 3.13 that
shouldn't be an issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 3.13)
 
 project(strawberry C CXX)
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
 if(APPLE)
   enable_language(OBJC OBJCXX)
 endif()


jonaski:
This is a development feature, and I don't think it should be hard-coded in the CMake file, what's the point of doing that?
`-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` can be passed to the CMake command to enable this for development, Qt creator already does this.

peterhoeg:
> what’s the point of doing that?

It’s simply to lower the barriers of entry for new developers. The only negative outcome is a few milliseconds of additional cmake execution time, so I would argue “why not do this?”

jonaski:
Passing `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` to CMake when needed should be sufficient.

PR #1958, opened 2026-01-12
Remove duplicate line

Perdu:
As [pointed by Copilot](https://github.com/strawberrymusicplayer/strawberry/pull/1835#discussion_r2659159253), this line is already present a few lines beforehand and can be removed.


Perdu:
CI error is an unrelated network error.

PR #2002, opened 2026-02-13
macOS/TagFetcher: improve Opus matching reliability (AcoustID POST + MusicBrainz fallback + safer fingerprinting)

OpsKonekt:
## Summary

This PR improves Tag Fetcher reliability for `.opus` files (notably files downloaded via yt-dlp) on macOS, while keeping the app self-contained (no external CLI dependency at runtime).

## Problem

Some files failed in Tag Fetcher with errors like:
- AcoustID HTTP 400 / invalid fingerprint
- "No decoded audio samples were produced by GStreamer"
- No useful fallback when AcoustID returned no recording IDs, especially when embedded tags were missing

### 1) AcoustID request robustness
- Switched lookup request from query-style GET to form POST (`application/x-www-form-urlencoded`)
- Improved API/network error parsing and diagnostics
- Kept request semantics identical (`format`, `client`, `duration`, `meta`, `fingerprint`)

### 2) MusicBrainz fallback when AcoustID fails
- Added metadata search fallback path in `TagFetcher`:
  - If AcoustID returns no MBIDs, perform a MusicBrainz recording search
- Added `MusicBrainzClient::StartSearchRequest(...)`
- Added query normalization/tokenization and result scoring (title/artist/album/duration heuristics)
- Preserved existing MBID path behavior

### 3) Better handling of missing tags
- If title/artist tags are missing, derive search metadata from filename
- Supports common filename patterns like track-number + artist + title

### 4) Better user diagnostics
- Track selection error page now shows technical details from backend failures
- Keeps details readable as plain text

### 5) Fingerprinting hardening + memory safety
- Hardened GStreamer decode path in `Chromaprinter`
- Added explicit in-memory PCM cap to avoid unbounded growth
- Added safer pad checks and clearer failure reasons

### 6) macOS bundle/runtime fixes
- Improved bundled GStreamer plugin discovery on startup
- Added plugin/runtime packaging adjustments needed for matroska/webm/opus flow
- Added `libgstriff` bundling when required by `libgstmatroska`
- Registry filename now includes executable timestamp to avoid stale cache reuse across rebuilds

## Validation

- Build: `cmake --build build-m1-full -j8` (passes)
- Manual tests on macOS Sonoma 14.6.1 (Apple Silicon):
  - Tag Fetcher with `.opus` files with and without embedded tags
  - AcoustID success path
  - MusicBrainz fallback path
  - Error diagnostics visibility in UI

## Commit breakdown

1. `musicbrainz: add metadata-search fallback when AcoustID fails`
2. `ui: show technical diagnostics in tag fetcher error page`
3. `chromaprinter: harden decoding and cap in-memory PCM usage`
4. `macos: bundle and discover GStreamer runtime dependencies reliably`

## Summary by CodeRabbit

* **New Features**
  * Added MusicBrainz search fallback when fingerprint-based identification cannot find a match.
  * Improved track matching using title, artist, album, and duration details.
  * Added technical error details to tag-fetching dialogs.

* **Bug Fixes**
  * Improved fingerprint generation reliability and validation.
  * Preserved clearer backend error messages for failed tag lookups.
  * Improved bundled GStreamer plugin detection and registry refresh behavior.

### DIFF
--- a/dist/macos/macgstcopy.sh
+++ b/dist/macos/macgstcopy.sh
@@ -45,6 +45,7 @@ fi
 
 mkdir -p "${bundledir}/Contents/PlugIns/gio-modules" || exit 1
 mkdir -p "${bundledir}/Contents/PlugIns/gstreamer" || exit 1
+mkdir -p "${bundledir}/Contents/Frameworks" || exit 1
 
 if [ -e "${GIO_EXTRA_MODULES}/libgiognutls.so" ]; then
   cp -v -f "${GIO_EXTRA_MODULES}/libgiognutls.so" "${bundledir}/Contents/PlugIns/gio-modules/" || exit 1
@@ -91,6 +92,8 @@ libgstid3demux
 libgstid3tag
 libgstisomp4
 libgstlame
+# Needed for WebM/Matroska containers that often wrap Opus streams.
+libgstmatroska
 libgstmpegpsdemux
 libgstmpegpsmux
 libgstmpegtsdemux
@@ -139,12 +142,20 @@ for gst_plugin in $gst_plugins; do
   fi
 done
 
+# libgstmatroska depends on libgstriff, which is not always pulled automatically.
+gst_library_path=$(dirname "${GST_PLUGIN_PATH}")
+if [ -e "${gst_library_path}/libgstriff-1.0.0.dylib" ]; then
+  cp -v -f "${gst_library_path}/libgstriff-1.0.0.dylib" "${bundledir}/Contents/Frameworks/" || exit 1
+  install_name_tool -id "@rpath/libgstriff-1.0.0.dylib" "${bundledir}/Contents/Frameworks/libgstriff-1.0.0.dylib"
+else
+  echo "Warning: Missing ${gst_library_path}/libgstriff-1.0.0.dylib."
+fi
+
 # libsoup is dynamically loaded by gstreamer, so it needs to be copied.
 if [ "${LIBSOUP_LIBRARY_PATH}" = "" ]; then
   echo "Warning: Set the LIBSOUP_LIBRARY_PATH environment variable for copying libsoup."
 else
   if [ -e "${LIBSOUP_LIBRARY_PATH}" ]; then
-    mkdir -p "${bundledir}/Contents/Frameworks" || exit 1
     cp -v -f "${LIBSOUP_LIBRARY_PATH}" "${bundledir}/Contents/Frameworks/" || exit 1
     install_name_tool -id "@rpath/$(basename ${LIBSOUP_LIBRARY_PATH})" "${bundledir}/Contents/Frameworks/$(basename ${LIBSOUP_LIBRARY_PATH})"
   else
--- a/src/dialogs/trackselectiondialog.h
+++ b/src/dialogs/trackselectiondialog.h
@@ -51,7 +51,7 @@ class TrackSelectionDialog : public QDialog {
 
  public Q_SLOTS:
   void FetchTagProgress(const Song &original_song, const QString &progress);
-  void FetchTagFinished(const Song &original_song, const SongList &songs_guessed);
+  void FetchTagFinished(const Song &original_song, const SongList &songs_guessed, const QString &error = QString());
 
   // QDialog
   void accept() override;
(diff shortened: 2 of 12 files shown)


.github/workflows/build.yml
jonaski: This commit shouldn't be included here.

dist/macos/macgstcopy.sh
```diff
@@ -91,6 +92,8 @@ libgstid3demux
 libgstid3tag
 libgstisomp4
 libgstlame
+# Needed for WebM/Matroska containers that often wrap Opus streams.
+libgstmatroska
```
jonaski, line 96: Matroska isn't currently built, it needs to be enabled in https://github.com/strawberrymusicplayer/strawberry-macos-dependencies/
how did you test this?

dist/macos/macgstcopy.sh
```diff
@@ -138,12 +141,20 @@ for gst_plugin in $gst_plugins; do
   fi
 done
 
+# libgstmatroska depends on libgstriff, which is not always pulled automatically.
```
jonaski, line 145: Then we need to investigate why, this seems like a workaround, isn't libgstriff-1.0.0.dylib linked?

src/dialogs/trackselectiondialog.h
```diff
@@ -78,6 +78,8 @@ class TrackSelectionDialog : public QDialog {
     Song original_song_;
     bool pending_;
     QString progress_string_;
+    // Carries technical diagnostics returned by TagFetcher for the error page.
```
jonaski, line 81: Don't think we need too many comments like this, since no other variables are commented.

src/engine/chromaprinter.cpp
```diff
@@ -214,17 +389,56 @@ GstFlowReturn Chromaprinter::NewBufferCallback(GstAppSink *app_sink, gpointer se
...
 
+  if (me->pcm_limit_reached_) {
+    // Ask upstream to finish once we have enough decoded material.
+    return GST_FLOW_EOS;
+  }
+
   return GST_FLOW_OK;
 
 }
-
```
jonaski, line 247: Avoid committing formatting changes in the same PR

src/engine/gststartup.cpp
```diff
@@ -145,7 +184,10 @@ void SetEnvironment() {
 #endif  // USE_BUNDLE
 
 #if defined(Q_OS_WIN32) || defined(Q_OS_MACOS)
-  QString gst_registry_filename = StandardPaths::WritableLocation(StandardPaths::StandardLocation::AppLocalDataLocation) + QStringLiteral("/gst-registry-%1-bin").arg(QCoreApplication::applicationVersion());
+  const QFileInfo executable_info(QCoreApplication::applicationFilePath());
+  const qint64 executable_stamp = executable_info.exists() ? executable_info.lastModified().toSecsSinceEpoch() : 0;
+  // Include executable timestamp to avoid reusing stale GStreamer registries across rebuilds.
```
jonaski, line 189: We only distribute a version once, so this shouldn't be necessary as far as I can tell. During development you can easily delete these manually. Is there a specific reason this is needed?

jonaski (changes requested):
Thanks, since there is a lot of changes here addressing different things, could you open a PR for each? This way it's easier to review.
I suggest to split into 5 different PR's similar to your commits:
1. Improved error handling 
2. Chromaprint pipeline changes
3. Additional MusicBrainz fallback to tag search
4. The gstreamer environmental variable changes (but I'm not sure we need these)
5. macgstcopy

You should be able to create 4 separate branches and cherry-pick the relevant commit to each branch.
Also, please rebase the commit messages to include capitalized "ClassName:" as prefix if the changes only affect one class, otherwise the commit title should be without a prefix.
And was a AI tool used for any of these changes?

jonaski:
> 
> ## Validation
> * Build: `cmake --build build-m1-full -j8` (passes)
> * Manual tests on macOS Sonoma 14.6.1 (Apple Silicon):
>   
>   * Tag Fetcher with `.opus` files with and without embedded tags
>   * AcoustID success path
>   * MusicBrainz fallback path
>   * Error diagnostics visibility in UI
> 

I'm confused by this, does this mean you tested everything?
And, is it tested on macOS only, or Linux and Windows too?

PR #2028, opened 2026-03-04
Grouped tracks are considered as a single track.

Vinzzzze:
consider the grouped tracks as one track so the queued tracks cannot interrupt the grouped tracks play :
- add a parameter to the Playlist::next_row method to handle this
- add a method Song::IsOnSameGrouping  
- add the attribute next_song_after_queued_ to remember the position where to go back when the queued rest is finish

add a configuration to define the behavior of the queued tracks in front of the grouped tracks :
- add the GROUPED_BEFORE_QUEUE_DEFAULT constexpr that define the default value for this configuration (default value set to 1 to keep the current behavior and ignore the grouped tracks)
- add the update_grouped_before_queue method to set the playlist objects with the value set in the ui
- use left_grouped_song_before_queue_ & init_grouped_song_before_queue_ to determine what to do next when we have queued tracks while reading grouped tracks
- add the ui in the backendsettingspage,in the background management tab.
- add a tool-tip on the created spinbox to explain how this parameter is used.
- translations had been prepared, but only the french one had been done.  

correct the previous row to navigate in the playlist :
- update the Playlist::previous_row code
- use only the navigation in the playlist to determine the previous row
- if we are on queued track read, use next_song_after_queued_ attribute to determine the position where to go back (that is to say that queued tracks are ignored in this navigation)

Values for the added parameter :
0 : wait for the end of all the grouped tracks on the same group than the current read track before playing the queued track(s)
1 : ignore the grouping attribute and read the queued track after the end of the current played track
n : wait for the end of the read of the n tracks of the current played track (the current one is the first of the n). If there is less than n grouped tracks to read, wait for the end of all the grouped tracks (same group than the current read track) to start the read of the queued track(s). If the read track is not a grouped track, the queued track(s) will be read just after it.
No maximum had been set.
I will use it this 0 : never interrupt grouped tracks.

## Summary by CodeRabbit

* **New Features**
  * Added a **“Grouping Before Queue”** option to playlist preferences, including a descriptive tooltip and a persisted numeric setting.

* **Improved Playback Behavior**
  * Enhanced grouped-track queue handling so queued playback returns/resumes more consistently within matching album/grouping contexts.
  * Improved “next song” state management when navigating tracks and when reshuffling or mutating playlists, preventing stale queued-next behavior.

### DIFF
--- a/src/playlist/playlist.cpp
+++ b/src/playlist/playlist.cpp
@@ -132,6 +132,7 @@ Playlist::Playlist(const SharedPtr<TaskManager> task_manager,
                    const int id,
                    const QString &special_type,
                    const bool favorite,
+                   const int grouped_before_queue,
                    QObject *parent)
     : QAbstractListModel(parent),
       is_loading_(false),
@@ -156,7 +157,10 @@ Playlist::Playlist(const SharedPtr<TaskManager> task_manager,
       scrobble_point_(-1),
       auto_sort_(false),
       sort_column_(Column::Title),
-      sort_order_(Qt::AscendingOrder) {
+      sort_order_(Qt::AscendingOrder),
+      left_grouped_song_before_queue_(grouped_before_queue),
+      init_grouped_song_before_queue_(grouped_before_queue),
+      next_song_after_queued_(-1) {
 
   undo_stack_->setUndoLimit(kUndoStackSize);
 
@@ -683,14 +687,72 @@ int Playlist::PreviousVirtualIndex(int i, const bool ignore_repeat_track) const 
 
 }
 
-int Playlist::next_row(const bool ignore_repeat_track) {
+int Playlist::next_row(const bool ignore_repeat_track, const bool no_grouping_track_count) {
 
-  // Any queued items take priority
-  if (!queue_->is_empty()) {
+  int next_virtual_index = next_song_after_queued_;
+
+  if (next_virtual_index < 0) {
+    next_virtual_index = NextVirtualIndex(current_virtual_index_, ignore_repeat_track);
+    switch (RepeatMode()) {
+      default:
+
+        // Compare with unsigned int to avoid to check if it is less than the count but positive
+        if (   static_cast<unsigned int>(current_virtual_index_) < static_cast<unsigned int>(virtual_items_.count())
+            && static_cast<unsigned int>(next_virtual_index) < static_cast<unsigned int>(virtual_items_.count())   ) {
+           // The two indexes are in the virtual items array
+           Song this_song = item_at(virtual_items_[current_virtual_index_])->EffectiveMetadata();
+           Song next_song = item_at(virtual_items_[next_virtual_index])->EffectiveMetadata();
+
+           if (   this_song.IsOnSameGrouping(next_song)
+               && (   left_grouped_song_before_queue_ == 0
+                   || no_grouping_track_count
+                   || --left_grouped_song_before_queue_ > 0   )   ) {
+              // We have two choices here :
+              // either resetting left_grouped_song_before_queue_ as soon as there is no queued song,
+              // that is too say, when a song will be queued, you will have to wait the setted number of grouped song before playing it
+              // either resetting left_grouped_song_before_queue_ when it reaches 0,
+              // that is too say, when a song will be queued, you will have to wait every n th grouped song before playing it
+              if (queue_->is_empty())
+              // if (left_grouped_song_before_queue_ == 0)
+              {
+                left_grouped_song_before_queue_ = init_grouped_song_before_queue_;
+              }
+              // The next song is on the same grouping than the current one,
+              // it cannot be interrupt by the queued items
+              break;
+           }
+        }
+
+        [[fallthrough]];
+
+      case PlaylistSequence::RepeatMode::Intro:
+      case PlaylistSequence::RepeatMode::Track:
+
+        // Reset the grouping values
+        left_grouped_song_before_queue_ = init_grouped_song_before_queue_;
+
+        // Any queued items take priority
+        if (!queue_->is_empty()) {
+          // Remember the location in the current playlist to go back right there after the queued play
+          next_song_after_queued_ = next_virtual_index;
+
+          // Play the queued song
+          return queue_->PeekNext();
+        }
+
+        break;
+    }
+  }
+  else if (!queue_->is_empty()) {
+    // We already remember the location where we want to go back after the queued play
+    // we just have to return the next queued song
     return queue_->PeekNext();
   }
+  else if (next_song_after_queued_ >= 0 && !no_grouping_track_count) {
+     // No more queued song, go back to the list to play
+     next_song_after_queued_ = -1;
+  }
 
-  int next_virtual_index = NextVirtualIndex(current_virtual_index_, ignore_repeat_track);
   if (next_virtual_index >= virtual_items_.count()) {
     // We've gone off the end of the playlist.
 
@@ -1054,6 +1116,9 @@ void Playlist::MoveItemsWithoutUndo(const QList<int> &source_rows, int pos) {
   PlaylistItemPtrList moved_items;
   moved_items.reserve(source_rows.count());
 
+  // Reset the item to be played after the queued, it is the safer behavior, if we are not on a dynamic playlist
+  if (!dynamic_playlist_) next_song_after_queued_ = -1;
+
   if (pos < 0) {
     pos = static_cast<int>(items_.count());
   }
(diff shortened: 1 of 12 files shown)


src/playlist/playlist.cpp
```diff
@@ -682,12 +743,16 @@ int Playlist::next_row(const bool ignore_repeat_track) {
 
 int Playlist::previous_row(const bool ignore_repeat_track) {
 
-  while (!played_indexes_.isEmpty()) {
-    const QPersistentModelIndex idx = played_indexes_.takeLast();
-    if (idx.isValid() && idx != current_item_index_) return idx.row();
```
jonaski, line 727: I believe removing this will break previous track feature as virtual_items_ is reshuffled

jonaski (changes requested):

src/playlist/playlist.cpp
Vinzzzze: Hello,

When writting this code I made some tests, I try to explain why I made theses changes, but it will be a little long.
When a track is changing, Mpris2::CurrentTrackChange is called. This call emits notifications and one of these notifications is CanGoPrevious.
CanGoPrevious call Playlist::previous_row, and previous_row is cleaning played_indexes_ when calling takeLast, so with the deleted code, played_indexes_ never own more than one element.
I try to use an iterator instead of takeLast, and using played_indexes_ to keep the played tracks, but with the code I deleted, where I replace takeLast by an iterator, I just switch between the last two tracks, so I remove completely this code to keep the last played_indexes_ in the list and being able to move back in the previous played tracks.
Besides, with the code before my updates, queued tracks change the history :
Let's say I have virtual indexes 1, 2, 3, 4, 5, 6. I can see it when I start at index 1, using next move to 2, then 3, then 4, then 5, then 6, and previous make me move from index 6, then 5, then 4, then 3, then 2 then 1, I can navigate like that as long as I don't enqueue any track.
When I enqueue tracks, let's say track A, the track sequence is lost. Let's say from the same sequence I enqueue track A after track 2, once I am on track A, if I go to the previous, I won't go back to track 2, if I go to the next, I won't go to track 3.
With my updates, if I enqueue tracks A, B, C after track 2, going next track 2, I go to track A, next I go to track B, previous I go to track 2, previous again I go to track 1, next I go to track C (the last remaining enqueued track), next I go to track 2, next I go to track 3, next I go to track 4, previous I go to track 3, previous I go to track 2, previous I go to track 1.
That is to say I keep the track sequence but forgot the enqueued track once they had been played.
I think that this is a good behavior, but may be I miss the point of your comment, or I am just wrong and this behavior is not the correct one.

jonaski: `Mpris2::CanGoPrevious()` clears `played_indexes_`, that's a bug, fix here: #2049
But with your changes, `played_indexes_` has no use, indexes are still appended in `Playlist::set_current_row`, but never used. It then only relies on `virtual_items_`, which will break since when `ReshuffleIndices` is called, that list is not consistent with the played indexes

src/engine/enginebase.h
```diff
@@ -236,6 +238,9 @@ class EngineBase : public QObject {
   bool channels_enabled_;
   int channels_;
 
+  // Grouping
+  int grouping_before_queue_;
```
jonaski: This shouldn't be in `EngineBase` has nothing to do with the engine.

src/core/application.cpp
```diff
@@ -142,7 +142,7 @@ class ApplicationImpl {
           app->MoveToThread(playlist_backend, database_->thread());
           return playlist_backend;
         }),
-        playlist_manager_([app]() { return new PlaylistManager(app->task_manager(), app->tagreader_client(), app->url_handlers(), app->playlist_backend(), app->collection_backend(), app->current_albumcover_loader()); }),
+        playlist_manager_([app]() { return new PlaylistManager(app->task_manager(), app->tagreader_client(), app->url_handlers(), app->playlist_backend(), app->collection_backend(), app->current_albumcover_loader(), GROUPED_BEFORE_QUEUE_DEFAULT); }),
```
jonaski: `GROUPED_BEFORE_QUEUE_DEFAULT` is only used in the Playlist manager, so passing this here seems uneccesarry.

src/constants/backendsettings.h
```diff
@@ -38,6 +38,7 @@ constexpr char kExclusiveMode[] = "exclusive_mode";
 constexpr char kVolumeControl[] = "volume_control";
 constexpr char kChannelsEnabled[] = "channels_enabled";
 constexpr char kChannels[] = "channels";
+constexpr char kGroupingBeforeQueue[] = "grouping_queue";
```
jonaski: This should be in playlist settings, not backend.

jonaski (changes requested):

jonaski:
I'm not sure if I want to add this, have to think about it. I doubt many users use shuffle by grouping and if we were to add something like this it makes sense to have the option for album shuffle too, not just grouping.
One thing I see: There is no need to bring the settings logic into `SmartPlaylistSearchPreview`, it only uses the Playlist model for view, no play logic is used there.

jonaski:
Another thing, please don't add translations updates into the PR's, it's just unnecessary noise, it's better to update the translation sources with `make update_translations` separately after the PR's are merged and let Crowin fill in the new tr's for all the languages.

Vinzzzze:
Hello,
I removed the translation files, then with the «make update_translations» I can regenerate the en_US file, and I have to create a PR with this file alone ?

For the SmartPlaylistSearchPreview, I use default parameter value. As it is a QObject derivated class, I left the Parent parameter at last position.

src/playlist/playlist.h
```diff
@@ -193,10 +197,12 @@ class Playlist : public QAbstractListModel {
   int last_played_row() const;
   void reset_last_played() { last_played_item_index_ = QPersistentModelIndex(); }
   void reset_played_indexes() { played_indexes_.clear(); }
-  int next_row(const bool ignore_repeat_track = false);
+  int next_row(const bool ignore_repeat_track = false, const bool no_grouping_track_count = true);
   int previous_row(const bool ignore_repeat_track = false) const;
   int take_previous_row(const bool ignore_repeat_track = false);
 
+  void update_grouped_before_queue(const int grouped_before_queue) { init_grouped_song_before_queue_ = grouped_before_queue; }
```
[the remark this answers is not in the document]
Vinzzzze: grouped_before_queue cannot be negative. This function is called when we reload the settings, and I don't want to reset the complete grouped setting.

src/playlist/playlist.cpp
```diff
@@ -641,14 +645,72 @@ int Playlist::PreviousVirtualIndex(int i, const bool ignore_repeat_track) const
 
 }
 
-int Playlist::next_row(const bool ignore_repeat_track) {
+int Playlist::next_row(const bool ignore_repeat_track, const bool no_grouping_track_count) {
 
-  // Any queued items take priority
-  if (!queue_->is_empty()) {
+  int next_virtual_index = next_song_after_queued_;
+
```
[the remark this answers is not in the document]
Vinzzzze: I am not sure there is a way for that : I have to think about it. Resetting next_song_after_queued_ can be wrong but it may be the safer way if the playlist is modified.

src/settings/playlistsettingspage.ui
```diff
@@ -98,6 +98,39 @@
...
+       <layout class="QHBoxLayout" name="layout_grouping_play_before_queue">
+        <item>
+         <widget class="QLabel" name="label_grouping_play_before_queue_1">
+          <property name="text">
+           <string>Number of grouped track played before play queued track</string>
+          </property>
+          <property name="indent">
+           <number>22</number>
+          </property>
+         </widget>
```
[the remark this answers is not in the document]
Vinzzzze: done

Vinzzzze:
Can you please confirm that you found the suggested refactor more readable ? For me, it is not the case, but if you agree I will do it.
The suggested refactor I am talking about is :

```
-         "<span style=\"font-weight:600;\">0:</span> "_L1 +
-         QObject::tr(" is used to say that the queued track will wait for the end of the current track group before being played.") +
+         QObject::tr("<span style=\"font-weight:600;\">0:</span> Play all tracks in the current group before queued tracks.") +
...
-         "<span style=\"font-weight:600;\">1:</span> "_L1 +
-         QObject::tr(" is used to say that the queued track will be played after the end of the current track, whatever it belongs to a group or not.") +
+         QObject::tr("<span style=\"font-weight:600;\">1:</span> Play queued tracks after the current track ends.")

```

PR #2029, opened 2026-03-04
Upgrade the MPris2 definitions

Vinzzzze:
Properties / methods updated :

- Tracks : return all the tracks in the current playlist
- GetTracksMetadata : return the specified tracklist metadata using Song::ToXesam method
- AddTracks : add the tracks to the current playlist
- RemoveTracks : remove the tracks from the current playlist
- GoTo : go to the specified track in the current playlist
- GetPlaylists : use the order parameter
- PlaylistItemsAdded : To know when a track is really added to a playlist
- PlaylistItemsRemoved : To know when a track is really removed from a playlist

### DIFF
--- a/src/mpris2/mpris2.cpp
+++ b/src/mpris2/mpris2.cpp
@@ -107,6 +107,10 @@ namespace mpris {
 constexpr char kMprisObjectPath[] = "/org/mpris/MediaPlayer2";
 constexpr char kServiceName[] = "org.mpris.MediaPlayer2.strawberry";
 constexpr char kFreedesktopPath[] = "org.freedesktop.DBus.Properties";
+constexpr char kTrackPrefix[] = "/org/strawberrymusicplayer/strawberry/Track/";
+constexpr char kNoTrack[] = "/org/mpris/MediaPlayer2/TrackList/NoTrack";
+
+constexpr int kNbTracksDisplaid = 20;
 
 Mpris2::Mpris2(const SharedPtr<Player> player,
                const SharedPtr<PlaylistManager> playlist_manager,
@@ -141,6 +145,8 @@ Mpris2::Mpris2(const SharedPtr<Player> player,
   QObject::connect(&*playlist_manager_, &PlaylistManager::PlaylistManagerInitialized, this, &Mpris2::PlaylistManagerInitialized);
   QObject::connect(&*playlist_manager_, &PlaylistManager::AllPlaylistsLoaded, this, &Mpris2::AllPlaylistsLoaded);
   QObject::connect(&*playlist_manager_, &PlaylistManager::CurrentSongChanged, this, &Mpris2::CurrentSongChanged);
+  QObject::connect(&*playlist_manager_, &PlaylistManager::PlaylistItemsAdded, this, &Mpris2::PlaylistItemsAdded);
+  QObject::connect(&*playlist_manager_, &PlaylistManager::PlaylistItemsRemoved, this, &Mpris2::PlaylistItemsRemoved);
   QObject::connect(&*playlist_manager_, &PlaylistManager::PlaylistChanged, this, &Mpris2::PlaylistChangedSlot);
   QObject::connect(&*playlist_manager_, &PlaylistManager::CurrentChanged, this, &Mpris2::PlaylistCollectionChanged);
 
@@ -340,6 +346,7 @@ QString Mpris2::LoopStatus() const {
   }
 
   switch (playlist_manager_->active() ? playlist_manager_->active()->RepeatMode() : playlist_manager_->sequence()->repeat_mode()) {
+    case PlaylistSequence::RepeatMode::Off: return u"None"_s;
     case PlaylistSequence::RepeatMode::Album:
     case PlaylistSequence::RepeatMode::Playlist: return u"Playlist"_s;
     case PlaylistSequence::RepeatMode::Track: return u"Track"_s;
@@ -350,8 +357,11 @@ QString Mpris2::LoopStatus() const {
 
 void Mpris2::SetLoopStatus(const QString &value) {
 
-  PlaylistSequence::RepeatMode mode = PlaylistSequence::RepeatMode::Off;
+  if (!playlist_manager_->sequence()) {
+    return;
+  }
 
+  PlaylistSequence::RepeatMode mode = PlaylistSequence::RepeatMode::Off;
   if (value == "None"_L1) {
     mode = PlaylistSequence::RepeatMode::Off;
   }
@@ -362,7 +372,12 @@ void Mpris2::SetLoopStatus(const QString &value) {
     mode = PlaylistSequence::RepeatMode::Playlist;
   }
 
-  playlist_manager_->active()->sequence()->SetRepeatMode(mode);
+  if (playlist_manager_->active()) {
+    playlist_manager_->active()->sequence()->SetRepeatMode(mode);
+  }
+  else {
+    playlist_manager_->sequence()->SetRepeatMode(mode);
+  }
 
 }
 
@@ -378,20 +393,38 @@ void Mpris2::SetRate(double rate) {
 
 bool Mpris2::Shuffle() const {
 
+  if (!playlist_manager_->active() && !playlist_manager_->sequence()) {
+    return false;
+  }
+
   const PlaylistSequence::ShuffleMode shuffle_mode = playlist_manager_->active() ? playlist_manager_->active()->ShuffleMode() : playlist_manager_->sequence()->shuffle_mode();
   return shuffle_mode != PlaylistSequence::ShuffleMode::Off;
 
 }
 
 void Mpris2::SetShuffle(bool enable) {
-  playlist_manager_->active()->sequence()->SetShuffleMode(enable ? PlaylistSequence::ShuffleMode::All : PlaylistSequence::ShuffleMode::Off);
+
+  if (!playlist_manager_->sequence()) {
+    return;
+  }
+
+  const PlaylistSequence::ShuffleMode mode = enable ? PlaylistSequence::ShuffleMode::All : PlaylistSequence::ShuffleMode::Off;
+  if (playlist_manager_->active()) {
+    playlist_manager_->active()->sequence()->SetShuffleMode(mode);
+  }
+  else {
+    playlist_manager_->sequence()->SetShuffleMode(mode);
+  }
+
 }
 
 QVariantMap Mpris2::Metadata() const { return last_metadata_; }
 
 double Mpris2::Rating() const {
-  float rating = playlist_manager_->active()->current_item_metadata().rating();
+
+  const float rating = playlist_manager_->active() ? playlist_manager_->active()->current_item_metadata().rating() : .0F;
   return (rating <= 0) ? 0 : rating;
+
 }
 
 void Mpris2::SetRating(double rating) {
@@ -408,11 +441,11 @@ void Mpris2::SetRating(double rating) {
 }
 
 int Mpris2::current_playlist_row() const {
-  return playlist_manager_->active()->current_row();
+  return playlist_manager_->active() ? playlist_manager_->active()->current_row() : -1;
 }
 
-QDBusObjectPath Mpris2::current_track_id(const int current_row) const {
-  return QDBusObjectPath(QStringLiteral("/org/strawberrymusicplayer/strawberry/Track/%1").arg(current_row));
+QDBusObjectPath Mpris2::current_track_id(const QUuid &current_row) const {
+  return QDBusObjectPath(QStringLiteral("/org/strawberrymusicplayer/strawberry/Track/%1").arg(current_row.toString(QUuid::WithoutBraces).replace(u'-', u'_')));
 }
 
 // We send Metadata change notification as soon as the process of changing song starts...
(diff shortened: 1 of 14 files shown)


jonaski:
> Create some new MPris properties / methods, my main worry is about the Shuffle property that was a boolean property and I migrated it to a string property. May be I should have created a new property. Properties / methods updated :

Shuffle needs to be bool: https://specifications.freedesktop.org/mpris/latest/Player_Interface.html#Property:Shuffle
And I'm not sure we should add methods that are not part of the official specifications either.

Vinzzzze:
ok, I will change my PR to match these requirements : no new method and Shuffle will be bool

src/mpris2/mpris2.cpp
```diff
@@ -50,10 +50,13 @@
 #include "core/song.h"
 #include "core/player.h"
 #include "engine/enginebase.h"
+#include "filterparser/filterparser.h"
```
jonaski: These includes aren't used anymore, remove

src/mpris2/mpris2.cpp
```diff
@@ -50,10 +50,13 @@
...
+#include "filterparser/filtertree.h"
 #include "playlist/playlist.h"
 #include "playlist/playlistitem.h"
 #include "playlist/playlistmanager.h"
 #include "playlist/playlistsequence.h"
+#include "queue/queue.h"
```
jonaski: Not used?

src/mpris2/mpris2.cpp
```diff
@@ -598,39 +603,94 @@ void Mpris2::OpenUri(const QString &uri) {
 }
 
 Track_Ids Mpris2::Tracks() const {
```
jonaski, line 673: Add newline after this line consistent with other functions

src/mpris2/mpris2.cpp
```diff
@@ -598,39 +603,94 @@ void Mpris2::OpenUri(const QString &uri) {
...
-  // TODO
-  return Track_Ids();
+  Track_Ids ret_value;
+
+  for (auto&& current_id : playlist_manager_->active()->virtual_items()) {
+    ret_value.push_back(current_track_id(current_id));
+  }
+  return ret_value;
```
jonaski: Add newline after this line consistent with other functions

src/mpris2/mpris2.cpp
```diff
@@ -598,39 +603,94 @@ void Mpris2::OpenUri(const QString &uri) {
...
 }
 
 bool Mpris2::CanEditTracks() const { return false; }
 
 TrackMetadata Mpris2::GetTracksMetadata(const Track_Ids &tracks) const {
```
jonaski, line 699: Add newline after this line consistent with other functions

src/mpris2/mpris2.cpp
```diff
@@ -598,39 +603,94 @@ void Mpris2::OpenUri(const QString &uri) {
...
+  TrackMetadata ret_value;
 
-  Q_UNUSED(tracks);
+  for (auto&& current_track_id : tracks) {
```
jonaski:
```suggestion
  for (auto &&current_track_id : tracks) {
```

src/mpris2/mpris2.cpp
```diff
@@ -598,39 +603,94 @@ void Mpris2::OpenUri(const QString &uri) {
...
+    if (split_path.isEmpty()) {
+      continue;
+    }
+    bool ok = false;
+    int track_id = split_path.last().toInt(&ok);
+    if (!ok) {
+      continue;
+    }
+    if (playlist_manager_->active()->has_item_at(track_id)) {
+      auto&& track_definition = playlist_manager_->active()->item_at(track_id);
```
jonaski:
```suggestion
      auto &&track_definition = playlist_manager_->active()->item_at(track_id);
```

src/mpris2/mpris2.cpp
```diff
@@ -598,39 +603,94 @@ void Mpris2::OpenUri(const QString &uri) {
 }
 
 Track_Ids Mpris2::Tracks() const {
-  // TODO
-  return Track_Ids();
+  Track_Ids ret_value;
+
+  for (auto&& current_id : playlist_manager_->active()->virtual_items()) {
```
jonaski:
```suggestion
  for (auto &&current_id : playlist_manager_->active()->virtual_items()) {
```

src/mpris2/mpris2.cpp
```diff
@@ -598,39 +603,94 @@ void Mpris2::OpenUri(const QString &uri) {
...
-  // TODO
+  if (split_path.isEmpty()) {
+    return;
+  }
+  bool ok = false;
+  int after_track_pos = split_path.last().toInt(&ok);
+  if (!ok) {
+    return;
+  }
+  QList<QUrl> listUrl;
```
jonaski: list_url

src/mpris2/mpris2.cpp
```diff
@@ -598,39 +603,94 @@ void Mpris2::OpenUri(const QString &uri) {
...
+
+  listUrl.push_back(QUrl(uri));
 
+  playlist_manager_->InsertUrls(playlist_manager_->current_id(),
```
jonaski: One line

src/mpris2/mpris2.cpp
```diff
@@ -598,39 +603,94 @@ void Mpris2::OpenUri(const QString &uri) {
...
+                                listUrl,
+                                after_track_pos,
+                                setAsCurrent,
+                                false);
 }
 
 void Mpris2::RemoveTrack(const QDBusObjectPath &trackId) {
```
jonaski, line 761: Newline after this

src/mpris2/mpris2.cpp
```diff
@@ -598,39 +603,94 @@ void Mpris2::OpenUri(const QString &uri) {
...
+
+  if (split_path.isEmpty()) {
+    return;
+  }
+  bool ok = false;
+  int track_id = split_path.last().toInt(&ok);
+  if (!ok) {
+    return;
+  }
+  playlist_manager_->active()->removeRows(track_id, 1);
```
jonaski: Newline after this

src/mpris2/mpris2.cpp
```diff
@@ -598,39 +603,94 @@ void Mpris2::OpenUri(const QString &uri) {
...
 }
 
 void Mpris2::GoTo(const QDBusObjectPath &trackId) {
```
jonaski, line 784: Newline after this

src/mpris2/mpris2.cpp
```diff
@@ -598,39 +603,94 @@ void Mpris2::OpenUri(const QString &uri) {
...
+
+  if (split_path.isEmpty()) {
+    return;
+  }
+  bool ok = false;
+  int track_id = split_path.last().toInt(&ok);
+  if (!ok) {
+    return;
+  }
+  playlist_manager_->active()->set_current_row(track_id);
```
jonaski: Newline after this

src/mpris2/mpris2.cpp
```diff
@@ -698,6 +755,32 @@ MprisPlaylistList Mpris2::GetPlaylists(quint32 index, quint32 max_count, const Q
     ret << mpris_playlist;
   }
 
+  if (!order.isEmpty()) {
+    if (order == "id"_L1) {
+      std::sort(ret.begin(),
```
jonaski: One line

src/mpris2/mpris2.cpp
```diff
@@ -698,6 +755,32 @@ MprisPlaylistList Mpris2::GetPlaylists(quint32 index, quint32 max_count, const Q
...
+                ret.end(),
+                [](MprisPlaylist& a,
+                   MprisPlaylist& b) {
+                  return a.id < b.id;
+                });
+    }
+    else if (order == "name"_L1) {
+      std::sort(ret.begin(),
```
jonaski: One line

src/mpris2/mpris2.cpp
```diff
@@ -698,6 +755,32 @@ MprisPlaylistList Mpris2::GetPlaylists(quint32 index, quint32 max_count, const Q
...
+                ret.end(),
+                [](MprisPlaylist& a,
+                   MprisPlaylist& b) {
+                  return a.name < b.name;
+                });
+    }
+    else if (order == "id"_L1) {
```
jonaski: This check is already done, remove?

src/playlist/playlist.h
```diff
@@ -190,6 +190,7 @@ class Playlist : public QAbstractListModel {
 
   int current_row() const;
   int last_played_row() const;
+  const QList<int>& virtual_items() const { return virtual_items_; }
```
jonaski:
```suggestion
  const QList<int> &virtual_items() const { return virtual_items_; }
```

jonaski (changes requested):

jonaski (approved):

src/mpris2/mpris2.cpp
```diff
@@ -594,43 +621,169 @@ void Mpris2::SetPosition(const QDBusObjectPath &trackId, qint64 offset) {
...
+      }
+    }
+  }
 
-  // TODO
+  const int insert_row = after_track_pos + 1;
+  playlist_manager_->InsertUrls(playlist_manager_->active_id(), QList<QUrl>() << QUrl(uri), insert_row, setAsCurrent, false);
+  if (playlist_manager_->active() && insert_row >= 0 && playlist_manager_->active()->has_item_at(insert_row)) {
+    Q_EMIT TrackAdded(GetTracksMetadata(Track_Ids() << QDBusObjectPath(QLatin1String(kTrackPrefix) + QString::number(insert_row))), afterTrack);
+  }
```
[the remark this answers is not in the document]
jonaski: This is a problem, need to look into this more closely when I have more time

src/mpris2/mpris2.cpp
```diff
@@ -698,6 +848,12 @@ MprisPlaylistList Mpris2::GetPlaylists(quint32 index, quint32 max_count, const Q
     ret << mpris_playlist;
   }
 
+  if (!order.isEmpty()) {
+    // Align with Orderings(): currently only "User" is advertised there.
+    if (order == "User"_L1) {
```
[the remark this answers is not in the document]
Vinzzzze: Initialy, copilot ask to use User instead of name and removing the sort on the id, I don't understand why it want to change again this point.

jonaski (changes requested):
According to the specifications https://mpris2.readthedocs.io/en/latest/mpris2.tracklist.html, the track ID's need to be unique and not change, when reshuffling, what songs `virtual_items_` point to will change so we can't use that, so I removed use of that so it's using the playlist item indexes, but I realize we can't use that either since it changes when inserting, removing or moving songs in the playlist.
We can't use song_id, because only collection songs have that, so I think we need to introduce a unique identifier to the playlist items.
Also, according to the documentation, TrackList should not return the entire playlist, but only the items in the queue + some of the previously played songs.
And AddTrack / RemoveTrack need to emit TrackAdded / TrackRemoved, but the signals needs to be emitted after the songs have been inserted in the playlist and a unique ID has been allocated.
So implementing this is much more complicated than I originally thought.

Vinzzzze:
> According to the specifications https://mpris2.readthedocs.io/en/latest/mpris2.tracklist.html, the track ID's need to be unique and not change, when reshuffling, what songs `virtual_items_` point to will change so we can't use that, so I removed use of that so it's using the playlist item indexes, but I realize we can't use that either since it changes when inserting, removing or moving songs in the playlist. We can't use song_id, because only collection songs have that, so I think we need to introduce a unique identifier to the playlist items. Also, according to the documentation, TrackList should not return the entire playlist, but only the items in the queue + some of the previously played songs. And AddTrack / RemoveTrack need to emit TrackAdded / TrackRemoved, but the signals needs to be emitted after the songs have been inserted in the playlist and a unique ID has been allocated. So implementing this is much more complicated than I originally thought.

Ok, got it. If you want, we can close this PR until having the right idea to solve theses issues. I am very interested by being able to control the player from the command line.
May be I should rather implement some new options in that purpose, or I can have a look to the MPris requirement and exchange with you on the solution I want to put in place before submitting a PR.

jonaski:
Let's just leave this open, then add the required changes separately, merge those first, then rebase this after it's implented.
We can implement a unique ID when new playlist items are inserted, which is saved to the playlist_items so it doesn't change when playlists are re-opened, I think QUuid: https://doc.qt.io/qt-6/quuid.html would be sufficient.
Then implement a function to `Playlist` which returns the previous 20 and next 20 ID's for use with MPRIS2 taking played indexes and next virtual items and queued items into consideration.
There also need to be a signal when items are inserted / removed back to MPRIS2 but only for the changes through MPRIS2 AddTrack / RemoveTrack, a bit unsure how difficult that would be to implement.

src/mpris2/mpris2.cpp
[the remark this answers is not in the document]
jonaski: Ignore this suggestion by Copilot, it needs to be implemented with signal slots from Playlist to MPRIS2.

Vinzzzze:
Hello,
I created the PR #2059 to add the unique identifier to items, this PR had to be validated first.
This PR is also included in this PR (that's why there are two commits, I will maintain them up-to-date) to show you the solution I imagine.
I think that now the code fulfill the mpris specifications.

src/mpris2/mpris2.cpp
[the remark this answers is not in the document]
Vinzzzze: I use the values from the MPris spec

jonaski:
I continued and merged this in https://github.com/strawberrymusicplayer/strawberry/pull/2140
I made the signal per playlist item, because `Playlist::removeRows` needs to know about it without a additional parameter to the overridden function.
Another thing I did was to add a QMap for mapping QUuid's to playlist items, the index can then easily be looked up using `items_.indexOf(item)` without iterating the whole playlist each time, avoids exposing the entire items to a public function too.
Thanks for your contributions.

PR #2036, opened 2026-03-14
Correct some translation in the parameter window

Vinzzzze:
Correct translations in the parameter window :
tab «Collection»
tab «Backend»


jonaski:
How do did update the translation source (`strawberry_en_US.ts`)?
Because it was still very out of sync with the codebase, I updated it now, see https://github.com/strawberrymusicplayer/strawberry/commit/a6224abd0356cafc73b42109a72195078e1b2454
I simply ran `make update_translations`, this hadn't been done by me in a very long time so it was completely out of sync

Vinzzzze:
ok, I will do this on each branch where I updated the translations.

PR #2056, opened 2026-04-06
Write MUSICBRAINZ_TRACKID to file tags on "Complete tags automatically"

VictorVow:
## Problem

\"Complete tags automatically\" retrieves the MusicBrainz recording UUID via AcoustID, uses it to query the MusicBrainz API for metadata, then silently discards it. The UUID never enters the `Result` struct and cannot reach the tag-write path. Only standard metadata (title, artist, album, year, track number) is written to files.

## Solution

Wire the recording UUID through the full pipeline so it is written to file tags alongside the other metadata.

### Changes

**`src/musicbrainz/musicbrainzclient.h` / `.cpp`**
- `Result` gains a `musicbrainz_recording_id_` field
- `MbIdRequestFinished` receives the `mbid` it looked up (already captured in the lambda) and stamps it onto every `Result` produced from that request

**`src/musicbrainz/tagfetcher.cpp`**
- `TagsFetched` copies `musicbrainz_recording_id_` onto the `Song` built from each `Result`

**`src/dialogs/trackselectiondialog.cpp`**
- `SaveData` includes `musicbrainz_recording_id` in the set of fields copied to the file on user confirmation

**`src/tagreader/tagreadertaglib.cpp`**
- ID3v2 (MP3, WAV, AIFF): TXXX frame with description `MUSICBRAINZ_TRACKID`
- Vorbis comments (FLAC, OGG, Opus): `MUSICBRAINZ_TRACKID` field
- APE tags (WavPack, Musepack, Monkey's Audio): `MUSICBRAINZ_TRACKID` item
- MP4/iTunes (M4A, AAC): `----:com.apple.iTunes:MusicBrainz Track Id`
- ASF/WMA: `MusicBrainz/Track Id` attribute
- Added TXXX read path for `MUSICBRAINZ_TRACKID` so tags written by this code are read back into the database on the next scan

## Notes

- All writes are guarded with `!isEmpty()` — no empty tags are written
- The existing UFID read path for `musicbrainz_recording_id` is preserved for files tagged by external tools (Picard, beets)
- The ASF write uses the existing `SetAsfAttribute` helper which already handles empty-string removal

### DIFF
--- a/src/dialogs/trackselectiondialog.cpp
+++ b/src/dialogs/trackselectiondialog.cpp
@@ -287,6 +287,7 @@ void TrackSelectionDialog::SaveData(const QList<Data> &_data) const {
     copy.set_album(new_metadata.album());
     copy.set_track(new_metadata.track());
     copy.set_year(new_metadata.year());
+    copy.set_musicbrainz_recording_id(new_metadata.musicbrainz_recording_id());
 
     const TagReaderResult result = tagreader_client_->WriteFileBlocking(copy.url().toLocalFile(), copy, TagReaderClient::SaveOption::Tags, SaveTagCoverData());
     if (!result.success()) {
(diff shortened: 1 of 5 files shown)


jonaski (approved):

jonaski:
Thanks!

PR #2059, opened 2026-04-10
Add the items Id to prepare the MPris update

Vinzzzze:
An Id had been added to the PlaylistItem using QUuid class
Events on adding / removing tracks in Playlist had been added
Boolean had been added to decide if the events on adding / removing had been added to methods


jonaski:
Done in https://github.com/strawberrymusicplayer/strawberry/pull/2140

PR #2138, opened 2026-06-11
CollectionModel: Eliminate multi-second UI freezes on large metadata update batches

### LINKED ISSUE #2112: Collection filter causes UI freeze during large Qobuz metadata batch updates
- [x] I have checked the [FAQ](https://wiki.strawberrymusicplayer.org/wiki/FAQ) for answers.
- [x] I have checked the [Changelog](https://github.com/strawberrymusicplayer/strawberry/blob/master/Changelog) that the issue is not already fixed.
- [x] I believe this issue is a bug, and not a general technical issue, question or feature request.

**Describe the bug**

When a collection filter (search expression) is active and Qobuz emits a large batch of `Song metadata changed` events, the main thread becomes unresponsive for many seconds — sometimes long enough that the window manager flags Strawberry as not responding.

A gdb stack trace of the stuck main thread shows the work happens entirely inside the collection filter, re-evaluating every active filter row for every batch-removed song:

```
Thread 1 (LWP 53339) "strawberry":
#0  QtPrivate::QMetaTypeForType<QString>::getDtor()::...
#1  QVariant::~QVariant()
#2  FilterTreeTerm::accept(Song const&) const
#3  FilterTreeAnd::accept(Song const&) const::{lambda}::operator()
...
#8  FilterTreeAnd::accept(Song const&) const
...
#14 FilterTreeOr::accept(Song const&) const
#15 CollectionFilter::filterAcceptsRow(int, QModelIndex const&) const
#16-19 (Qt model internals)
#20 QAbstractItemModel::endRemoveRows()
#21 CollectionModel::RemoveSongsInternal(QList<Song> const&)
#22 CollectionModel::ProcessUpdate()
#23-29 (Qt signal/slot dispatch from a QTimer)
```

Resource state while stuck: RSS ~600 MB (stable, no leak), main thread at ~50% CPU continuously, all worker threads idle in `g_cond_wait` / `pthread_cond_timedwait`. So this is a busy main thread, not a deadlock.

The debug log shows 3300 `CollectionModel: Song metadata for … changed` lines in one ~20-minute session, with bursts of 130–160 changes in a single millisecond. Each burst lines up with the rebatched `ProcessUpdate()` call that the stack trace was caught in. All affected songs in the visible bursts are classical compilation tracks (Berliner Philharmoniker, Schönberg piano works, etc.) that exist in the local library and presumably matched against Qobuz.

**To Reproduce**

1. Have a large local collection (a few thousand songs) including classical/compilation albums that match Qobuz catalog entries.
2. Enable Qobuz and let it complete its genre-metadata fetch.
3. Apply a collection filter that matches very few songs (e.g. type `Heino` into the collection search field).
4. While the filter is active, Qobuz emits a batch of metadata updates → UI hangs.

**Expected behavior**

Collection filter evaluation during model updates should be bounded. Either: (a) coalesce `ProcessUpdate()` into fewer `beginRemoveRows`/`endRemoveRows` ranges so the filter only re-runs once per batch, or (b) skip filter re-evaluation for rows that did not change identity, or (c) move filter evaluation off the UI thread.

**System Information:**
- Operating system: CachyOS Linux (kernel 7.0.10-1-cachyos)
- Strawberry Version: 1.2.19-66-g44079d18 (current master, locally built)
- Qt: 6.11.1
- GStreamer: system default

**Additional context**

The batch trigger appears to be the genre-metadata-fetch path added in `ea629aed` ("Get genre metadata for Tidal, Qobuz and Spotify"), but the underlying issue is in `CollectionModel::RemoveSongsInternal` / `CollectionFilter::filterAcceptsRow` — any large metadata batch with an active filter should be enough to reproduce it. Related closed issues: #1482 (slow collection search) and #1162 (search-as-you-type unpleasant on large collections) — both touch the same code path but were closed without a fix to the filter–batch interaction.

mzilinski:
Fixes #2112.

## Problem

When a streaming service (or any backend operation) delivers a large metadata batch, the collection tree freezes for multiple seconds. A batch of N changed songs fans out in `CollectionModel` into thousands of scattered remove/re-add pairs, each emitting its own `begin/endRemoveRows` / `begin/endInsertRows` pair.

This is the freeze reported in #2112. I filed that issue blaming the active collection filter — the measurements below show that diagnosis was wrong: the filter is innocent, the freeze is view-driven and happens with or without a filter. Same symptom, corrected root cause.

## Root cause (measured)

Benchmarked with the real `CollectionFilter` proxy and an expanded `QTreeView` attached, on a 6000-song library with 2000 scattered changed songs (every 3rd song moves to a new album, i.e. the remove/re-add path). The benchmark harness is posted as a comment below so the numbers can be reproduced:

| Variant | Time |
|---|---|
| Current master (no batching) | 1647 ms |
| Contiguous-run batching only (first two commits) | 1765 ms (within run-to-run noise of master — neither helps nor hurts this case) |
| Single transaction above threshold (this PR, complete) | **33 ms** |

The hang is **view-driven, not filter-driven**: with and without an active filter the numbers are identical (1655 ms vs 1688 ms), and without any view attached even unbatched processing takes only ~42 ms. The cost is QTreeView rebuilding its internal `viewItems` bookkeeping on every `rows{Inserted,Removed}` signal — O(n) per signal, O(n²) for the batch on an expanded tree. The existing code comment blaming `QSortFilterProxyModel::filterAcceptsRow` turned out to be misleading; the per-row signals are the problem, regardless of filtering.

## What the commits do

1. **Batch contiguous row removals** per parent into single `begin/endRemoveRows` pairs. Does not help the scattered case above, but collapses whole-album removals into one signal pair.
2. **Batch `beginInsertRows` per container** in `AddSongsInternal`.
3. **Collapse bulk-sized batches into one transaction.** Above a threshold of 1000 pending songs (counted across the update queue — the queue is an artifact of 400-song chunking), all per-row signals are suppressed and the whole drain becomes a single transaction. This also fixes a real bug: `AddReAddOrUpdateSongsInternal` re-queued derived changes instead of applying them inline, so a bulk transaction wrapped no tree mutation at all and the real work landed in a second transaction one tick later (an empty plus a double relayout).
4. **Use `layoutAboutToBeChanged`/`layoutChanged` instead of a model reset** for that transaction, remapping persistent indexes across the mutation. Unlike `begin/endResetModel`, this keeps view state alive: selection, scroll position and expanded containers survive — including a selected song that is removed and re-added under a new container, because remapping is identity-based (song id / container level+key / divider key) rather than pointer-based. Pointer-based matching would not be safe: the heap can hand a new item the address of one just deleted in the same drain.
5. **Tests** for all of the above.

A `Reset` queued behind a bulk batch is never folded into the transaction (it carries its own `begin/endResetModel` and raises `loading_`); the drain stops when one surfaces and handles it on the next tick.

## Verification

- New tests cover: per-row signal suppression above the threshold, the per-row path below it, exactly one layout change per bulk drain (no reset), balanced signal pairs when a `Reset` is queued mid-batch, persistent-index remapping at the model level, and an end-to-end test where a real `QTreeView` attached through `CollectionFilter` keeps its expanded containers, its selection (including the re-added song) and its scroll position across a bulk update.
- Negative control: with the bulk branch temporarily reverted to `begin/endResetModel`, both preservation tests fail (all persistent indexes invalidated, view state lost) — they measure the actual behavior, not implementation details.
- Full `collectionmodel_test` suite: 17/17 green, rebased onto current master. (Run directly: `QT_QPA_PLATFORM=offscreen ./tests/collectionmodel_test` from the build dir.)

## Notes for review

- The 1000-song threshold is a heuristic and deliberately **not** gated on an active filter (the hang is filter-independent). Happy to tune.
- Scroll position is "not thrown back to the top" (what the reset did) — content above the viewport can still shift by the row delta of the mutation.
- The layout-change transaction costs the same as the reset it replaces (33 ms vs 34 ms in the benchmark).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Improvements**
  * Speed up large collection refreshes by batching updates more efficiently and reducing unnecessary UI signaling.
  * Preserve view state (expanded/collapsed containers, selection, and scroll position) across bulk changes, including in-place metadata edits and remove/re-add scenarios.
* **Bug Fixes**
  * Album artwork now more reliably shows the “no cover” icon when cover art isn’t available.
* **Tests**
  * Added/extended regression tests to verify bulk-update behavior, expected signal patterns, and persistent index/view state preservation.

### DIFF
--- a/src/collection/collectionmodel.cpp
+++ b/src/collection/collectionmodel.cpp
@@ -38,6 +38,7 @@
 #include <QIODevice>
 #include <QList>
 #include <QSet>
+#include <QHash>
 #include <QMap>
 #include <QMetaType>
 #include <QVariant>
@@ -61,6 +62,7 @@
 #include "core/iconloader.h"
 #include "core/settings.h"
 #include "core/songmimedata.h"
+#include "core/scopedflag.h"
 #include "collectionfilteroptions.h"
 #include "collectionquery.h"
 #include "collectionbackend.h"
@@ -96,6 +98,7 @@ CollectionModel::CollectionModel(const SharedPtr<CollectionBackend> backend, con
       total_artist_count_(0),
       total_album_count_(0),
       loading_(false),
+      bulk_mode_(false),
       icon_disk_cache_(new QNetworkDiskCache(this)) {
 
   setObjectName(backend_->source() == Song::Source::Collection ? QLatin1String(QObject::metaObject()->className()) : QStringLiteral("%1%2").arg(Song::DescriptionForSource(backend_->source()), QLatin1String(QObject::metaObject()->className())));
@@ -451,12 +454,6 @@ void CollectionModel::ScheduleAddSongs(const SongList &songs) {
 
 }
 
-void CollectionModel::ScheduleUpdateSongs(const SongList &songs) {
-
-  ScheduleUpdate(CollectionModelUpdate::Type::Update, songs);
-
-}
-
 void CollectionModel::ScheduleRemoveSongs(const SongList &songs) {
 
   ScheduleUpdate(CollectionModelUpdate::Type::Remove, songs);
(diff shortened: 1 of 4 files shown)


mzilinski:
For reproducing the benchmark numbers in the description — this is the harness, written as an extra test in `tests/src/collectionmodel_test.cpp` on this branch (it reuses the fixture's `MakeSongs`/`Drain` helpers). Not committed, since it logs timings instead of asserting behavior:

<details>
<summary>Benchmark test (click to expand)</summary>

```cpp
TEST_F(CollectionModelTest, ScatteredUpdateBenchmark) {

  AddSong(u"seed"_s, u"seed"_s, u"seed"_s, 1);
  backend_->AddOrUpdateSongs(MakeSongs(6000, u"bench"_s));
  Drain();

  QTreeView view;
  view.setModel(collection_filter_);
  view.show();
  view.expandAll();
  QCoreApplication::processEvents();

  SongList stored;
  const QList<CollectionItem*> nodes = model_->song_nodes();
  for (CollectionItem *node : nodes) {
    if (node->metadata.title().startsWith(u"bench_"_s)) stored << node->metadata;
  }
  ASSERT_EQ(6000, stored.count());
  std::sort(stored.begin(), stored.end(), [](const Song &a, const Song &b) { return a.id() < b.id(); });

  // Every 3rd song moves to a new album: 2000 scattered remove/re-add pairs.
  SongList changed;
  for (int i = 0; i < stored.count(); i += 3) {
    Song s = stored[i];
    s.set_album(s.album() + u"_moved"_s);
    changed << s;
  }
  ASSERT_EQ(2000, changed.count());

  qint64 transaction_ms = -1;
  QElapsedTimer t;
  QObject::connect(&*model_, &CollectionModel::layoutAboutToBeChanged, &view, [&t]() { t.start(); });
  QObject::connect(&*model_, &CollectionModel::layoutChanged, &view, [&t, &transaction_ms]() { transaction_ms = t.elapsed(); });

  QElapsedTimer total;
  total.start();
  backend_->AddOrUpdateSongs(changed);
  Drain();
  qWarning() << "SCATTERED: layout transaction" << transaction_ms << "ms, total incl drain idle" << total.elapsed() << "ms";

  EXPECT_EQ(6001, static_cast<int>(model_->song_nodes().count()));

}
```

</details>

Notes:

- Needs `#include <QTreeView>` in the test file. Run with `QT_QPA_PLATFORM=offscreen ./tests/collectionmodel_test --gtest_filter='*ScatteredUpdateBenchmark*'`.
- On this branch the transaction window is measured between `layoutAboutToBeChanged` and the end of `layoutChanged` delivery (proxy rebuild and view relayout happen inside the emit, and these lambdas are connected last, so they bracket the full reaction).
- On master (or the intermediate commits) those signals don't fire — there, time `Drain()` instead and subtract its fixed 250 ms idle window; per-row signal storms dominate the remainder. That is how the 1647/1765 ms numbers were obtained.
- "Total incl drain idle" also contains the 250 ms idle window, 20 ms timer ticks and the SQL UPDATEs for 2000 songs; the user-visible main-thread stall is the transaction number.

*Edited: the original version of this harness used a `ScopedPrettyCoversOff` settings guard; that workaround was removed in review (`AlbumIcon` now null-checks the cover loader, see 8ee456f9), so the harness no longer needs a settings override.*

src/collection/collectionmodel.cpp
```diff
@@ -81,6 +82,19 @@ const int CollectionModel::kPrettyCoverSize = 32;
 namespace {
 constexpr char kPixmapDiskCacheDir[] = "pixmapcache";
 constexpr char kVariousArtists[] = QT_TR_NOOP("Various artists");
+
+// Sets a bool for the duration of a scope and clears it on the way out, even on
+// early return. Used to keep bulk_mode_ from getting stuck on if a dispatch ever
+// returns early inside a bulk layout-change transaction.
+class ScopedFlag {
```
jonaski: Move this class to it's own file in Core so it can be used in other places as well.

src/collection/collectionmodel.cpp
```diff
@@ -81,6 +82,19 @@ const int CollectionModel::kPrettyCoverSize = 32;
 namespace {
 constexpr char kPixmapDiskCacheDir[] = "pixmapcache";
 constexpr char kVariousArtists[] = QT_TR_NOOP("Various artists");
+
+// Sets a bool for the duration of a scope and clears it on the way out, even on
```
jonaski: In this code, longer lines are used, so reduce newlines for the comments as otherwise they take up a a lot of space, and preferably split on comma or punctuation. That's the style used in the codebase.

tests/src/collectionmodel_test.cpp
```diff
@@ -47,6 +59,34 @@ using std::make_shared;
...
+// for the duration of a test and restores the previous setting on the way
+// out, even on early ASSERT returns.
+class ScopedPrettyCoversOff {
+ public:
+  ScopedPrettyCoversOff() {
+    Settings s;
+    s.beginGroup(CollectionSettings::kSettingsGroup);
+    had_value_ = s.contains(CollectionSettings::kPrettyCovers);
+    old_value_ = s.value(CollectionSettings::kPrettyCovers);
+    s.setValue(CollectionSettings::kPrettyCovers, false);
```
jonaski: Don't modify the config in unit tests, instead add a null check in `CollectionModel::AlbumIcon`:
`if (!albumcover_loader_ || !item) return pixmap_no_cover_;`
Then you can get rid of all this.

jonaski (changes requested):

mzilinski:
Thank you for the review! All three points are addressed:

- `ScopedFlag` now lives in its own file in core (`src/core/scopedflag.h`) — 94cb4230
- The comments are reformatted to longer lines, split on punctuation — 2182779c
- `AlbumIcon` now checks for a null cover loader and the tests no longer modify the configuration; the `ScopedPrettyCoversOff` workaround is gone — 8ee456f9. I also updated the benchmark comment above to match.

Thanks again for taking the time to review this, there is other stuff coming, some improvement to the radio.
Have a great day!

jonaski:
@coderabbitai full review

src/collection/collectionmodel.cpp
```diff
@@ -470,12 +467,122 @@ void CollectionModel::ProcessUpdate() {
...
+  // Above this many pending songs we drain the queued run in one layout-change transaction, suppressing the per-row begin/end{Insert,Remove}Rows (and dataChanged) signals.
+  // Without this, every per-row signal makes an attached QTreeView rebuild its internal item bookkeeping (and a QSortFilterProxyModel re-evaluate filterAcceptsRow);
+  // when a large streaming-service metadata batch fans out into thousands of remove/add pairs the cost is O(rows) per signal and the UI freezes.
+  // A single layoutAboutToBeChanged/layoutChanged pair collapses it into one relayout, and unlike a model reset it keeps view state alive:
+  // persistent indexes are remapped by item identity below, so selection, scroll position and expanded containers survive.
+  // The hang is view-driven, not filter-driven, so this is deliberately NOT gated on an active filter. The value is a tunable heuristic.
+  constexpr int kBulkUpdateSongThreshold = 1000;
+
+  int pending_songs = 0;
+  for (const CollectionModelUpdate &update : updates_) {
```
jonaski:
```suggestion
  for (const CollectionModelUpdate &update : std::as_const(updates_)) {
```

jonaski: Also, I think this loop need to break if there is a `CollectionModelUpdate::Type::Reset`, and not count anything after a reset in pending_songs

jonaski (changes requested):

mzilinski:
Thanks for the second look! Both points are addressed in f6b9ba11:

- The count loop now uses `std::as_const(updates_)` to avoid the QQueue detach (matching the `identities` loop below).
- It also `break`s on the first `CollectionModelUpdate::Type::Reset`, so songs queued behind a Reset are no longer counted toward the bulk threshold — they belong to a later tick, since the bulk drain stops at the Reset anyway. This avoids spinning up a full layout-change transaction for a handful of songs that merely sit in front of a large post-Reset batch.

Builds clean and all 17 CollectionModelTest cases still pass. Ready for another look whenever you have time — thanks again!

jonaski (approved):

PR #2148, opened 2026-06-15
Prevent duplicate songs with identical URL

### LINKED ISSUE #1341: Strawberry creating duplicate tracks when using "move to collection"
This appears to be the same error as seen in issue 902, but that one's closed and I have a lot more details, so I opened this new ticket. If you'd prefer to re-open that issue and have me move this information in, let me know. 

-In short: When I hit "move to collection" some tracks appear duplicated in the "Collection" tab. Goes away after I "organize" the album.

I made a youtube video showing the error in realtime if that's easier. 

https://www.youtube.com/watch?v=CQlU7elMxTE

Details:

-Fully updated Arch Linux
-Strawberry 1.0.22 (although it's been happening for a while for me, last few versions)

-Happens with .opus and .ogg files at minimum, I believe it's happened with MP3 in the past as well
-If it happens to one track, will happen to all subsequent tracks in the album
-Anecdotally seems to usually happen starting at track 10 or 11. Sometimes later, ~but don't believe I've ever seen it earlier than track 10, making me think something about the double digit renaming causes it~.
-Happens almost every album for me, so I'm guessing it's related to something in my workflow/auto track renaming.
-Hitting "organize" will usually fix it, but not always. I've even seen it duplicate the files again, causing 4 instances of the same track, but that is rarer and I cannot consistently reproduce it.

jonaski:
Fixes #1341

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate song records by deduplicating on `(url, beginning)` and reusing existing entries when rescan items arrive without database IDs.
  * Improved rescans and directory scanning by NFC-normalizing paths, avoiding false “deleted” matches and correctly restoring unavailable tracks.
* **Tests**
  * Added regression coverage for duplicate-URL updates, directory healing on rediscovery, preservation of existing play/rating stats, and separation of same-URL multi-part tracks.
* **Chores**
  * Updated the database schema to version 23 and expanded bundled schema resources.

### DIFF
--- a/data/data.qrc
+++ b/data/data.qrc
@@ -14,6 +14,7 @@
         <file>schema/schema-20.sql</file>
         <file>schema/schema-21.sql</file>
         <file>schema/schema-22.sql</file>
+        <file>schema/schema-23.sql</file>
         <file>schema/device-schema.sql</file>
         <file>style/strawberry.css</file>
         <file>style/smartplaylistsearchterm.css</file>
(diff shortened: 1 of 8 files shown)


jonaski:
@coderabbitai review

PR #2217, opened 2026-07-08
Song: Set metadata from stream even if empty

brhartman:
When streaming music, for the title, artist, and album tags, if one song has the metadata set and the next song in the same stream does not, the previous song's metadata was carried over.  For example, if song A had the album set, then song B did not, song B would keep song A's album, which would cause issues for lyric and cover look ups.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved metadata refresh for radio/streamed tracks: title, artist, and album can now update correctly even when incoming values are empty, as long as the track is identified as radio/streamed.
  * Stream/radio details are more reliably kept in sync with the latest available metadata source, reducing stale or mismatched display information.


jonaski (approved):
Makes sense, thanks!

PR #2255, opened 2026-07-26
GstEnginePipeline: Abandon pipelines stuck in a state change

mzilinski:
## Problem

While playing an internet radio stream on macOS, the TCP connection died silently (network change; `lsof` showed no open connections left for the process). The souphttpsrc streaming task ended up blocked forever in `gst_soup_http_src_do_request()` waiting on a `GCond` that `unlock()` failed to signal — despite souphttpsrc's default `timeout=15`/`retries=3`. This is an upstream bug in gst-plugins-good/libsoup (GStreamer 1.28.2), reported with full thread samples of two occurrences in https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/5238.

The consequence on Strawberry's side: `gst_element_set_state(pipeline, GST_STATE_NULL)` blocked forever in `gst_pad_stop_task()` waiting for the streaming task to exit. The pipeline never emitted `Finished()`, so it was parked in `old_pipelines_` forever, and as soon as its last reference was dropped, `~GstEnginePipeline()` blocked forever in `waitForFinished()` — freezing playback control and eventually the whole application. A thread sample of the hung process (attached to the upstream report) shows two pipelines stuck this way in the same session: the app retried with a fresh pipeline, which hung identically.

## Fix

Strawberry cannot fix the upstream race, but it can contain the damage:

- `Finish()` now starts a 20 second watchdog timer. If the NULL transition does not settle in time, the pipeline is marked stuck and `Finished()` is force-emitted so `GstEngine` can move on and playback keeps working.
- The destructor's wait for pending state changes is now deadline-bounded. A stuck pipeline is deliberately leaked (no `gst_element_set_state(NULL)`, no unref), since a GStreamer thread may still be inside `gst_element_set_state()` holding locks on it — a leaked pipeline object and one dead pool thread are the lesser harm compared to a frozen application.
- The duplicated Finished-emission logic is factored into `EmitFinishedOnce()`, which also stops the watchdog on the normal path.

## Testing

- All 16 unit tests pass (`ctest`, excluding `lyrics_live_tests`).
- Normal path exercised in the real app against a local HTTP stream server: play → stop (fade-out → `Finishing pipeline` → NULL → `Finished` → destructor) over three pipeline lifecycles, plus quitting via SIGTERM while a stream was playing. No watchdog false positives; every pipeline was destroyed cleanly and without delay.
- The stuck path itself could not be reproduced locally (the upstream race needs a silently dead connection; with a local stalling server souphttpsrc's timeout aborts correctly), so that path is verified by review only. Worst case if the watchdog misfires is a leaked pipeline plus an error log entry instead of a freeze.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CM8kS4UmXUiG6nDaM8rTeA

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented application freezes when media pipelines become stuck during shutdown.
  * Added timeouts for stalled pipeline shutdowns, including unresponsive network sources.
  * Improved handling of incomplete state transitions so shutdown can continue without waiting indefinitely.
  * Prevented callbacks from accessing pipelines after they have been closed.
  * Ensured pipeline completion notifications are emitted only once.
  * Improved cleanup behavior for pipelines that finish shutting down normally.
  * Improved recovery when abandoning pipelines that cannot be stopped cleanly.

### DIFF
--- a/src/engine/gstenginepipeline.cpp
+++ b/src/engine/gstenginepipeline.cpp
@@ -51,6 +51,7 @@
 #include <QFutureWatcher>
 #include <QMutex>
 #include <QMutexLocker>
+#include <QDeadlineTimer>
 #include <QMetaType>
 #include <QByteArray>
 #include <QList>
@@ -93,6 +94,10 @@ constexpr int GST_PLAY_FLAG_SOFT_VOLUME = 0x00000010;
 constexpr int kGstStateTimeoutNanosecs = 10000000;
 constexpr std::chrono::milliseconds kFaderFudgeMsec = 2000ms;
 constexpr std::chrono::milliseconds kFaderTimeoutMsec = 3000ms;
+
+// How long a requested state change may run before the pipeline is considered stuck and abandoned.
+// A buggy source element can block gst_element_set_state() forever (e.g. souphttpsrc waiting on a HTTP request that unlock() fails to cancel when the connection silently died); without a deadline that would freeze pipeline teardown and eventually the whole application.
+constexpr std::chrono::seconds kFinishWatchdogTimeout = 20s;
 
 constexpr int kEqBandCount = 10;
 constexpr int kEqBandFrequencies[] = { 60, 170, 310, 600, 1000, 3000, 6000, 12000, 14000, 16000 };
@@ -218,6 +223,8 @@ GstEnginePipeline::GstEnginePipeline(QObject *parent)
       about_to_finish_(false),
       finish_requested_(false),
       finished_(false),
+      stuck_(false),
+      timer_finish_watchdog_(new QTimer(this)),
       bus_message_generation_(0),
       set_state_async_in_progress_(0) {
 
@@ -236,6 +243,10 @@ GstEnginePipeline::GstEnginePipeline(QObject *parent)
   timer_fader_timeout_->setSingleShot(true);
   QObject::connect(timer_fader_timeout_, &QTimer::timeout, this, &GstEnginePipeline::FaderTimelineTimeout);
 
+  timer_finish_watchdog_->setSingleShot(true);
+  timer_finish_watchdog_->setInterval(kFinishWatchdogTimeout);
+  QObject::connect(timer_finish_watchdog_, &QTimer::timeout, this, &GstEnginePipeline::FinishWatchdogTimeout);
+
 }
 
 GstEnginePipeline::~GstEnginePipeline() {
@@ -246,8 +257,9 @@ GstEnginePipeline::~GstEnginePipeline() {
 
     // Wait for any ongoing state changes for this pipeline to complete before setting to NULL.
     // This prevents race conditions with async state transitions.
-    {
-      // Copy futures to local list to avoid holding mutex during waitForFinished()
+    // The wait is deadline-bounded: a state change stuck inside a broken source element would otherwise block this destructor (and with it the calling thread) forever.
+    if (!stuck_.load()) {
+      // Copy futures to local list to avoid holding mutex while waiting
       QList<QFuture<GstStateChangeReturn>> futures_to_wait;
       {
         QMutexLocker locker(&mutex_pending_state_changes_);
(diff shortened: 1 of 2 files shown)


mzilinski:
Rebased onto current master. One thing worth calling out from the conflict resolution: the destructor's stuck path now also skips the new wait for manufactured-EOS pad sends from 2d0511d99. Rationale: in the stuck case the NULL flush never happened, so a pad send blocked in a serialized event would never be released and `waitForFinished()` could block forever — the same class of hang this PR is meant to contain. On the normal path the ordering from 2d0511d99 is preserved unchanged.

I'm also running this patch as my daily driver now; if the upstream souphttpsrc race triggers again I'll report back whether the watchdog caught it.

jonaski:
This is not new, and I think there are more blocking cases than this, but I'm not a fan of silently leaking anything, I'd rather address the problem upstream.

mzilinski:
I agree the real fix belongs upstream — the root cause is a race in souphttpsrc where the streaming thread can wait forever in `gst_soup_http_src_do_request()` when the TCP connection dies silently, and `unlock()` doesn't wake it (diagnosed from thread samples, reproduced twice).

I did try to take it upstream: I filed gst-plugins-good issue #5238 with the analysis, but it was closed because I had drafted it with AI assistance and the GStreamer project doesn't accept AI-generated reports. And even once a proper upstream fix exists, it will take time to land and reach users through distro/Homebrew updates.

On the leak: it only happens in a case that is already lost — the state change to NULL has hung, so the alternative isn't a clean teardown, it's the destructor blocking forever and freezing the whole application on exit. The watchdog logs a warning when it triggers, so it's at least not silent in the log. I'm open to making it louder if you'd prefer.

You mentioned there are more blocking cases than this one — do you have an idea for a more general solution that would cover those as well? I'm happy to rework this PR in that direction, or keep it as a narrow interim defense until an upstream fix exists and is deployed.

jonaski:
I'm not a fan of silently ignoring a real bug, it makes it a lot harder to actually identify and fix the real bug.
As for GStreamer, I wouldn't hold my breath on anything getting fixed unless we do it our-self, there are over 2200 issues open.

mzilinski:
You're right that the watchdog only contains the damage — so I took your "unless we do it our-self" seriously and wrote the upstream fix. It's done, and I can now demonstrate both the bug and the fix deterministically.

**Root cause, briefly:** souphttpsrc's streaming thread dispatches send/read operations to the soup session thread and then waits on `session_cond` in an unbounded `g_cond_wait()` loop whose *only* wake-up is the libsoup completion callback. `unlock()` cancels the `GCancellable` and broadcasts `retry.cond` — it never touches `session_cond`. The design therefore depends entirely on libsoup delivering the cancellation. When a TCP connection dies silently, it sometimes doesn't (the race I sampled in gstreamer#5238); the callback never runs and the streaming thread sleeps forever. That's what blocks `gst_element_set_state(NULL)` in `gst_pad_stop_task()` and freezes the application. The closures the callback writes into are stack-allocated in the streaming thread — which is *why* the code has no choice but to wait forever: an early exit would leave the callback a dangling pointer.

**The fix** (patch + repro harness: https://gist.github.com/mzilinski/59dedf03d3b14ebe3857e3718b2ab4c8):

1. `unlock()`/`stop()` additionally broadcast `session_cond`, and the wait loops also check `g_cancellable_is_cancelled()` — cancellation always wakes the streaming thread, regardless of what libsoup does.
2. The send/read closures become refcounted and shared with the session thread, so the streaming thread can abandon an operation whose callback hasn't run; a late-firing callback drops its result harmlessly.
3. While a read is in flight the closure owns the mapped output buffer (an abandoned read can't scribble over memory returned to a downstream pool); an abandoned send clears `src->msg` since the message may still be in flight.

208 insertions, 70 deletions, one file. Builds warning-free against 1.28.6 and current git main (the file is identical in both).

**Verification.** The network race itself can't be triggered on demand, but its mechanical state — "the completion callback never runs" — can: freeze the libsoup session thread via `PTRACE_ATTACH` while an icecast stream plays, then stop the pipeline. Results on GStreamer 1.28.6:

| | behaviour after stop with the session thread frozen |
|---|---|
| stock souphttpsrc | hangs indefinitely: streaming thread in `g_cond_wait` inside `read_buffer()`, main thread in `gst_pad_stop_task()` — the exact stacks from my macOS samples in #5238. Notably `timeout=15` never fires either: its timer lives on that same thread, which is why the property didn't help in the field. |
| patched souphttpsrc | `unlock()` wakes the streaming thread **21 µs** later (`interrupted before the read completed` in the debug log) and teardown proceeds. (Under this harness it later parks in `stream_clear()` waiting for a cleanup idle — an artifact of freezing the *whole* thread, which is strictly stronger than the real bug, where the session thread stays alive and only one callback is lost. In the real-bug condition teardown completes.) |
| reqwesthttpsrc | clean full exit ~1 s after stop even with **all** tokio workers frozen — its cancellation (`AbortHandle`) needs no cooperation from the I/O threads at all. |

The last row is why I'm confident the patch has the right shape: it moves souphttpsrc to the same cancellation model the Rust element already uses.

For transparency: like the original diagnosis, the patch and the measurements were developed with AI assistance. That's exactly why the harness exists — every claim here can be re-verified from scratch in a few minutes without taking anything on trust, and I've re-run the harness myself (patched wake-up 85 µs after `unlock()` on my run).

**The Rust element as a practical alternative.** Since souphttpsrc is only reachable through element selection, there's a second way to make the hang class structurally impossible without touching GStreamer at all: prefer `reqwesthttpsrc` from gst-plugins-rs. I validated it in Strawberry with `GST_PLUGIN_FEATURE_RANK=reqwesthttpsrc:MAX` against Qobuz for a full session — 8 pipeline generations, rapid seeks (each 40–110 ms including re-buffer), gapless track changes, every pipeline torn down cleanly in 350–550 ms, no errors. It works without any code change: `SourceSetupCallback` is already feature-detected, `user-agent` and the proxy properties exist on it; `ssl-strict`/`automatic-redirect` don't, but reqwest follows redirects itself and neither matters for the streaming services.

If we wanted to offer this to users, the shape I'd suggest is a small opt-in behind a hidden setting, following the existing rank-adjustment pattern in `GstStartup::Initialize()` — affected users install the distro package (`gst-plugin-reqwest` exists in Arch/Fedora/openSUSE repos) and set one key. Concretely (compile-tested; runtime effect identical to the env var I ran the Qobuz session with):

<details>
<summary>proposed diff for src/engine/gststartup.cpp</summary>

```diff
--- a/src/engine/gststartup.cpp
+++ b/src/engine/gststartup.cpp
@@ -32,6 +32,8 @@

 #include "core/logging.h"
 #include "core/standardpaths.h"
+#include "core/settings.h"
+#include "constants/backendsettings.h"
 #include "utilities/envutils.h"

 #ifdef HAVE_GSTFASTSPECTRUM
@@ -75,6 +77,25 @@ void Initialize() {
   }
 #endif

+  {
+    Settings s;
+    s.beginGroup(BackendSettings::kSettingsGroup);
+    const bool prefer_reqwesthttpsrc = s.value("prefer_reqwesthttpsrc", false).toBool();
+    s.endGroup();
+    if (prefer_reqwesthttpsrc) {
+      // Opt-in: prefer the Rust HTTP source element over souphttpsrc.
+      // souphttpsrc can hang forever in a state change when a connection dies silently (https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/5238); reqwesthttpsrc's cancellation does not depend on the HTTP library delivering it, which makes that hang class structurally impossible.
+      if (GstPluginFeature *reqwesthttpsrc = gst_registry_lookup_feature(gst_registry_get(), "reqwesthttpsrc")) {
+        qLog(Info) << "Preferring reqwesthttpsrc over souphttpsrc for HTTP streams";
+        gst_plugin_feature_set_rank(reqwesthttpsrc, GST_RANK_PRIMARY + 1);
+        gst_object_unref(reqwesthttpsrc);
+      }
+      else {
+        qLog(Error) << "prefer_reqwesthttpsrc is set, but the reqwesthttpsrc element is not available (install gst-plugins-rs / gst-plugin-reqwest)";
+      }
+    }
+  }
+
 }

 void SetEnvironment() {
```

</details>

Honest caveats: for the bundled builds (macOS/Windows) it would mean shipping a Rust-built plugin via macgstcopy.sh, which is a real packaging cost; the element lacks `ssl-strict` (irrelevant for the streaming services, but a behavior difference for users with broken certs); and in my Qobuz session I saw short buffering start/finished cycles (~50 ms every few seconds) in the log that I haven't yet compared against souphttpsrc on the same stream — nothing audible was reported, but I'd want that A/B before making it a recommendation rather than an escape hatch.

**On actually landing this upstream:** my own route there is compromised — as I mentioned, #5238 was closed over AI-assisted drafting, so I don't expect an MR from me to be received neutrally. My realistic path is re-authoring the patch by hand and starting over via Discourse, which I'm prepared to do, but it will take a while. In the meantime, the analysis, the repro harness, the prototype patch and all measurements are free for anyone with a cleaner path to an MR than mine — no strings and no authorship attached; the fix landing is all I care about. The harness makes the bug and the fix independently verifiable in a few minutes, with no need to trust my analysis or its provenance.

**For this PR** that leaves three options, in my order of preference:

1. Keep the watchdog as a clearly-labelled containment (error log pointing at the upstream bug — nothing silent) until a fix reaches distro/Homebrew releases. Given the upstream path is slow at best, this is what actually helps users in the meantime.
2. Additionally/instead, the opt-in reqwesthttpsrc rank bump described above — no bundled-build changes needed if it's gated on the element being present, so distro users get an escape hatch at zero packaging cost.
3. Close this PR and wait for upstream alone.

Happy to go whichever way you prefer.

PR #2282, opened 2026-08-19
Add option to save current playlist columns as default

bespsm:
Adds a "Save current columns as default" entry to the playlist header's right-click menu, next to "Reset columns to default".

It saves the current column order, widths, visibility, sort indicator and per-column text alignment to settings. "Reset columns to default" then restores this saved layout instead of falling back to the built-in default column layout, if one was saved. Behavior is unchanged when no default has been saved.

## Summary by CodeRabbit

* **New Features**
  * Added an option to save the current playlist column layout as the default.
  * Resetting playlist columns now restores the saved layout when available.
  * Saved column alignments are preserved and restored automatically.
  * Reset continues to use the built-in layout when no compatible saved default exists.

### DIFF
--- a/src/constants/playlistsettings.h
+++ b/src/constants/playlistsettings.h
@@ -56,6 +56,10 @@ constexpr char kState[] = "state";
 constexpr char kColumnAlignments[] = "column_alignments";
 constexpr char kRatingLocked[] = "rating_locked";
 
+constexpr char kSavedDefaultStateVersion[] = "saved_default_state_version";
+constexpr char kSavedDefaultState[] = "saved_default_state";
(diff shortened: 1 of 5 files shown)


jonaski:
I don't quite understand the need for this. The primary reason I made the reset column to default was because someone might mess up the columns completely, or there is bug introducing a stuck state. But if introducing this it will be possible to save a broken column state as the default with no way to restore except manually editing the config.

bespsm:
@jonaski many time I spotted the issue, where the playlist columns were suddenly reverted to default way and I had to sort it back, I cannot reproduce it... buts eems like you solved this issue.. I'll keep in in draft if you don't midn

bespsm:
or just close it and come back if it's still relevant

